PR #1, opened 2018-08-25
Make it build on Linux

captnfab:
- LHOS File functions
Added missing file functions for POSIX.

- Includes for C/C++std functions
Added missing includes for C/C++std functions so that gcc can build the project.


handsomematt:
thanks for the pull request, Linux compataiblity is definietely something I want to 100% mantain so it's a huge help 😄

PR #2, opened 2018-11-06
Suggested patch

captnfab:
Hi @handsomematt,
I see you've done some work on openblack.
Trying to build it on Linux gave me a couple of warnings and errors.
Here is a pull request where each commit is related to one of those problems, and I hope those shouldn't break anything on non-linux platforms.
I can split the PR if there are some changes you don't want to apply.


handsomematt:
Whoops looks like Visual Studio doesn't refactor class names in #ifdef trees.

Thank you for your changes as always. 😄

captnfab:
You're welcome, thanks for your work :)

PR #5, opened 2019-03-02
Fixes for building with GCC and working on Linux

captnfab:
- MapMaterial refers both to a type (struct) and a variable (member of struct Country). I had to rename it to compile with gcc. It is probably a good practice to do so.
- Some c functions were used without proper includes
- The use of backslash in paths is only compatible with windows, so I added an `#ifdef` for Unices


handsomematt:
Thank you! 😄 I'll try and get some Path/File related classes down to avoid cross platform issues in the future.

captnfab:
You're welcome.
I think Windows accepts / as a separator for relative paths, this could be a solution too: changing working directory once with absolute paths in a OS-friendly way, and then always use relative paths using / as a separator?

PR #17, opened 2019-08-20
script-validation - Validate the parameters given to a script command

raffclar:


src/LHScriptX/CommandSignature.h
```diff
@@ -44,8 +45,13 @@ class ScriptCommandParameter
...
 	ScriptCommandParameter(ParameterType type = ParameterType::None):
 	    _type(type) {}
-	ScriptCommandParameter(const std::string& value):
-	    _type(ParameterType::String) { SetString(value); }
+	ScriptCommandParameter(const std::string value):
+	    _type(ParameterType::String)
+	{
+		// TODO: Hacky, avoid new
+		_value._string = new std::string();
+		*_value._string = value;
```
raffclar: Any advice on this is welcome. Before the fix it was crashing the process as there is nothing allocated for value.

PR #21, opened 2019-08-31
allow user to define the CXX_STANDARD

NeroBurner:
Allow the user to set the C++ standard to allow to use even newer
standards


handsomematt:
Is there a way to make this require a minimum of 17? We use a lot of features from it

NeroBurner:
I think if the user requires a version less then the default the build will break and it is the users fault

handsomematt:
makes sense, thank you 👍

PR #23, opened 2019-08-31
use snprintf because iota is not ansi C

NeroBurner:
got error
```
/home/nero/repos/openblack/src/3D/MeshPack.cpp: In member function ‘void OpenBlack::MeshPack::loadTextures(OpenBlack::File&)’:
/home/nero/repos/openblack/src/3D/MeshPack.cpp:162:3: error: ‘itoa’ was not declared in this scope
  162 |   itoa(tex.first, sBlockID, 16);
      |   ^~~~
```

```
g++ --version
gcc (GCC) 9.1.0
```

alternative (and info that iota is not ansi C from here:
https://stackoverflow.com/a/6462973

Edit: got a CI job with the error, exciting :)
https://travis-ci.org/NeroBurner/openblack/jobs/579247589

### DIFF
--- a/src/3D/MeshPack.cpp
+++ b/src/3D/MeshPack.cpp
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <stdexcept>
 #include <stdint.h>
+#include <stdlib.h> // snprintf
 
 namespace OpenBlack
 {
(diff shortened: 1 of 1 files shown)


handsomematt:
seems good, would rather a solution involving std::string like here: http://www.strudel.org.uk/itoa/ in the future

NeroBurner:
we could use `std::stringstream` with `std::hex` formatter
https://stackoverflow.com/a/25144137

PR #26, opened 2019-08-31
Compilation fixes

janisozaur:


janisozaur:
Ah, didn't notice @NeroBurner already submitted those fixes moments ago.

PR #27, opened 2019-08-31
Logging

raffclar:
image


handsomematt:
Code is fine, just not sure if I was right to advise the dependency was added as a submodule. Would Hunter be able to fetch this @NeroBurner ?

NeroBurner:
yes there is a spdlog package already in hunter
https://docs.hunter.sh/en/latest/packages/pkg/spdlog.html#index-0

raffclar:
@handsomematt Alright if we merge this once I rebase it?

PR #28, opened 2019-08-31
Add travis build script

raffclar:


linux_build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+info() {
+    echo -e "${BG}$1${NC}";
+}
+
+mkdir -p build
+cd build
+cmake -DCMAKE_BUILD_TYPE=Debug ..
+
+FLAGS="-j8 -Wall"
```
janisozaur: Why does this have to be a variable, when it's only used in a single place?

Travis only offers up to 2 CPUs, setting 8 jobs will only cause thrashing. Why not use ninja instead?

`-Wall` seems like a compiler option, not make one, what is the purpose of this?

.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: cpp
+dist: bionic
+compiler: gcc
+
+script:
+- mkdir -p build
+- cd build
+- cmake -DCMAKE_BUILD_TYPE=Debug ..
+- make CXXFLAGS="-j2 -Wall"
```
janisozaur: I doubt `CXXFLAGS` will be honoured here, you should set it at cmake configuration step instead.

handsomematt:
prefer #35

PR #29, opened 2019-08-31
Explicitly link to stdc++fs

janisozaur:


janisozaur:
GCC prior to version 9 requires explicit linking. I'm not sure if that's the only problem with such older compilers that warrant

> GCC 9.1.0+ is required due to bugs in earlier versions with C++17 std.

note in the readme, so I haven't removed it. We (OpenRCT2, OpenLoco) use C++17 with GCC7 without any issues

PR #33, opened 2019-08-31
Fix include guards

janisozaur:
A large portion of the project used double include guards. There were
more `#pragma`s than the traditional ones, so I guessed that was the
original intention.


raffclar:
👍

handsomematt:
perfect, thank you!

PR #35, opened 2019-09-01
Ci and dependency manager

NeroBurner:
as I've seen another proposal for a [travis build script](https://github.com/openblack/openblack/pull/28) I thought I'd chime in and propose mine too

This one uses docker container to select the distribution to build for. Currently Ubuntu 19.04 (disco dingo) is used, but it can be changed to other distros/releases as well

I'm also in the progress to add Hunter as dependency manager to use it for the appveyor Visual Studio builds (and maybe for the mingw-cross compilation builds)


janisozaur:
It seems to be building all the libraries (glm, flew, sdl2…) inside CI? A few times? At least for mingw toolchain. That's quite pointless and takes a lot of time. I'm against doing such useless work.

janisozaur:
I'd also like to learn the reasons why is there a requirement for Ubuntu 19.04? Why not use Bionic/18.04 LTS for now in addition to something like Arch that has current versions of compilers and libraries? Or at least Fedora, which is generally much more recent than any deb-based one.

handsomematt:
I think Ubuntu 19.04 was just what I was testing on locally, obviously lowering the dependency would be good. As long as we can use GCC9.1 as there was a hard bug with I think C++17 std::filesystem feature.

handsomematt:
Also I'm not sure of the advantages to using Docker for building? Can it not just use matrixes?

NeroBurner:
You can use Docker on your machine to check a failing build. Matrixes can be used only on travis

Hunter supports cache servers for binary builds: https://docs.hunter.sh/en/latest/user-guides/hunter-user/github-cache-server.html?highlight=cache

or we could use travis caching feature to store the `.hunter` directory in cache

janisozaur:
You were not linking to libstdc++fs earlier, but that was addressed in https://github.com/openblack/openblack/pull/29

We use c++17 and its filesystem with GCC7 without any problems

NeroBurner:
tried Ubuntu bionic, but it has cmake 3.10 and minimum reqired seems to be 3.12. Can we use a lower minimum cmake version?

NeroBurner:
`#include <filesystem>` not found on Ubuntu 18.04
https://travis-ci.org/NeroBurner/openblack/jobs/579416772

janisozaur:
@NeroBurner see https://github.com/openblack/openblack/pull/36 for addressing the GCC7 compat

NeroBurner:
@janisozaur works! Nice!
https://travis-ci.org/NeroBurner/openblack/jobs/579421302

src/Game.cpp
```diff
@@ -453,10 +458,21 @@ void Game::LoadLandscape(const std::string& name)
...
+void Game::SetGamePath(const std::string &gamePath)
 {
-	static std::string sGamePath;
+	if (gamePath.empty())
+	{
+		return;
+	}
+	if (!fs::exists(gamePath))
+	{
+		return;
```
janisozaur: Since we were passed some non-null value here and perform a check, it would be nice to report this mishap to the user.

PR #36, opened 2019-09-01
Add support for GCC7's experimental filesystem

janisozaur:
The one that ships with ubuntu 18.04 by default

### DIFF
--- a/src/Common/File.h
+++ b/src/Common/File.h
@@ -20,11 +20,13 @@
 
 #pragma once
 
-#include <filesystem>
+#include <experimental/filesystem>
 #include <string>
 #include <cstdio>
 #include <cassert>
 #include <unordered_map>
+
+namespace fs = std::experimental::filesystem;
 
 namespace OpenBlack
 {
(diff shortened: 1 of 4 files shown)


janisozaur:
screenFetch-2019-09-01_14-21-43

The only issue is the cmake packaged is only in version 3.10, but apparently I've solved this over a year ago:
https://launchpad.net/~janisozaur/+archive/ubuntu/cmake-update-bionic

No other changes were necessary, works just fine.

raffclar:
Why not just use Kitware's PPA?

janisozaur:
I'd do that, if I had known about that. Can you post a link?

raffclar:
https://github.com/openblack/openblack/blob/bbab129e587e747f2ad559ffff2a015ebfae4f34/.travis.yml

janisozaur:
Ah, in that case I haven't used it because it's not a PPA (by which I mean the archives backed by https://launchpad.org which have tooling inside ubuntu).

But using an external repo is fine too, it's just a few more manual steps.

janisozaur:
https://github.com/openblack/openblack/pull/35#issuecomment-526913855 tries to lower the required version to 3.10.

src/Common/File.h
```diff
@@ -20,12 +20,14 @@
 
 #pragma once
 
-#include <filesystem>
+#include <experimental/filesystem>
```
janisozaur: This could be wrapped inside `__has_include` check verifying presence of non-experimental FS, but this should work as-is in most toolchains. We use experimental/boost FS like so in OpenLoco: https://github.com/OpenLoco/OpenLoco/blob/master/src/openloco/platform/platform.h

handsomematt:
excellent, much better then the hard requirement on GCC9.1

PR #40, opened 2019-09-01
Fixed glew on wayland

NeroBurner:
Glew can't open glx display when it's running on wayland session and thus
returns an error. But glXGetProcAddress is fully usable on wayland, so we
can just ignore the "no glx display" error.

Note that GLEW_ERROR_NO_GLX_DISPLAY needs glew >= 2.1. Older versions assume
that glx display is always available and will just crash.

source: https://github.com/SuperTux/supertux/commit/06976f8ed70c7137807b7b312c07f5698832be9a


NeroBurner:
It works under Wayland! (KDE Plasma KWin). Just set the environment Variable `SDL_VIDEODRIVER=wayland`

Screenshot_20190901_180312

PR #42, opened 2019-09-01
add option -g to set gamePath

NeroBurner:


src/Game.cpp
```diff
@@ -453,10 +458,21 @@ void Game::LoadLandscape(const std::string& name)
...
+void Game::SetGamePath(const std::string &gamePath)
 {
-	static std::string sGamePath;
+	if (gamePath.empty())
+	{
+		return;
+	}
+	if (!fs::exists(gamePath))
+	{
+		return;
```
janisozaur: Since we were passed some non-null value here and perform a check, it would be nice to report this mishap to the user.

NeroBurner: Indeed! Added a log message

PR #53, opened 2019-09-03
hunter: manually provide EnTT and use it

NeroBurner:
EnTT is a good cmake project with no dependencies, easy to download with hunter


handsomematt:
Can you remove the submodule for EnTT as well as the `add_subdirectory("${EXTERNAL_DIR}/entt")` line too since this resolves any need for that 👍

NeroBurner:
Ubuntu has no libentt-dev as far as I can see. This would break the Ubuntu builds

handsomematt:
makes sense, not sure what to do about that then 🤔

NeroBurner:
I propose to keep the entt submodule :man_shrugging:

PR #54, opened 2019-09-03
hunter: also provide spdlog

NeroBurner:
hunter can also provide spdlog

we should think a bit how we want the dependencies to be provided. The dependencies tend to get a bit messy


handsomematt:
same comment as #53 - remove the submodule as this resolves the need for it

PR #56, opened 2019-09-03
Ci appveyor cache

NeroBurner:
enable cache, also includes https://github.com/openblack/openblack/pull/55


handsomematt:
I think it should be: `- C:\.hunter -> CMakeLists.txt` or similar, that way .hunter cache is invalidated if we change CMakeLists.txt (which would probably be to change dependencies)

NeroBurner:
CMakeLists changes much to frequently in my opinion. I'd say manual cache invalidation should be enough. Maybe add a file at `ci/appveyor.cache` and make it dependent on that file

PR #58, opened 2019-09-03
update imgui to 1.72b, use submodule and cmake target

NeroBurner:
- use submodule for imgui instead of sources directly
- crate a cmake target to link to openblack


NeroBurner:
Ubuntu 18.04 fails because cmake is older than 3.11 which introduced

> You can also finally directly add INTERFACE targets to IMPORTED libraries
https://cliutils.gitlab.io/modern-cmake/chapters/intro/newcmake.html

will have to work around that

handsomematt:
#59 had these changes

PR #62, opened 2019-09-04
LLVMViewer: fix some warnings

NeroBurner:
- don't use char* directly as formatter string, feed it into a format
- use size_t for selected variable


handsomematt:
ty

PR #71, opened 2019-09-05
let window be resizeable: set flag SDL_WINDOW_RESIZABLE

NeroBurner:


NeroBurner:
rebased on current master @handsomematt

PR #74, opened 2019-09-06
Update travis to use new docker images

### LINKED ISSUE #66: Prebuild docker images
Instead of building docker image for each of the build jobs, build them once and reuse. This is much faster and avoids eating lots of cycles.

Right now it of about 11 minutes spent in https://travis-ci.com/openblack/openblack/jobs/231148513, 560 seconds are spent preparing docker image, while just 75s actually building the project.

In OpenRCT2, a far bigger project, we download docker images prepared on docker hub then build the project inside utilising ccache. For openblack this could bring the total build time down to (guesstimated) 1-2 minutes, or about 85% faster.

This would also let reuse the system images to obtain reproducible builds, addressing https://github.com/openblack/openblack/issues/20 properly.

janisozaur:
Images on docker hub are still going through (hopefully final) build and this is what's needed here to make use of them.

I will retrigger the CI tomorrow, once the images are all done.

Fixes #66

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,30 +23,24 @@ matrix:
       os: linux
       env: >
         BUILD_FLAVOR=ubuntu
-        BUILD_RELEASE=bionic
-        BUILD_ARCH=amd64
-        BUILD_PACKAGES="g++ ca-certificates"
         TOOLCHAIN=gcc-cxx17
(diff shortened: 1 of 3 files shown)


handsomematt:
fantastic, thank you!

PR #78, opened 2019-09-07
Rename Mesh to Meshes

raffclar:


handsomematt:
I think `MeshID` is the most descriptive name, e.g: `MeshID::BuildingCeltic1` that way it just specifies it's an ID and not an actual reference to a mesh.

PR #83, opened 2019-09-07
Add Fedora rawhide travis build job

janisozaur:


janisozaur:
@NeroBurner wayland gets detected by SDL2 and its define gets enabled, pulling in https://github.com/openblack/openblack/blob/9433fdcf90e79edad9472e7545d4b621636babc1/src/GameWindow.cpp#L27 but `wayland-egl` doesn't get linked. Will you take a look?

janisozaur:
Hmm, the CI job passed, but didn't work the way I expected it when testing locally. Nevermind then.

PR #93, opened 2019-09-10
GameWindow: MULTISAMPLE specific recreate of SDL_Window

NeroBurner:
Once again on Wayland the SDL_Window can't be created when
SDL_GL_MULTISAMPLEBUFFERS and SDL_GL_MULTISAMPLESAMPLES are set. On error
reset those values and try to one again create the SDL_Window without
those variables set.


handsomematt:
Improve on this a bit in 615218bdfb60ffdabba23fbc3a32347b7ef8904e

Now it tries 4 samples first, then 2, then 1 and finally disables it all together - this is still not the best solution as it removes user preference, but it is a start.

PR #95, opened 2019-09-10
Vertex attribs refact

bwrsandman:
Refactoring VertexDecl mainly.

The constructor has been greatly simplified to an api that looks a lot like bgfx.
Remove a lot of explicit `sizeofs` that are mostly opengl specific.
Moved the `VertexDecl` to vertex buffer.
Converted all usages of vertex attributes to this api.
Force alignment on 4 bytes in terrain.

Fix a few issues.
This includes #94

### DIFF
--- a/src/3D/L3DSubMesh.cpp
+++ b/src/3D/L3DSubMesh.cpp
@@ -121,10 +121,11 @@ void L3DSubMesh::Load(IStream& stream)
 		primIndexOffset += primitive.numTriangles * 3;
 	}
 
-	VertexDecl decl(3);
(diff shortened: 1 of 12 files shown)


handsomematt:
Are vertex attribs bound to a vertex buffer object instead of a vertex array object?

bwrsandman:
On OpenGL they are part of the VAO. They only need to be set once.
On other APIs there is no concept of VAO.
This is why the binding of attributes is called from mesh.

handsomematt:
Sounds good to me

PR #87, opened 2019-09-09
Rendering refact

bwrsandman:
Small changes done while preparing for bgfx.

Small changes to gitignore and cmake.

Fixed an sdl bug

Removed some dead code.

Improve readability of Gui class and prepare it for bgfx-imgui.

PR #96, opened 2019-09-10
Texture refact

bwrsandman:
Texture changes from #87 but without the private constructor.


handsomematt:
Building failing on windows, need a `#include <stdint.h>` in Texture2D.h I think

janisozaur:
This is c++, so `cstdint`

PR #97, opened 2019-09-10
OpenGL clean-up in headers

bwrsandman:
Add enums for opengl types.
Remove include of opengl from headers and move them to impl files.

### DIFF
--- a/src/Graphics/Mesh.cpp
+++ b/src/Graphics/Mesh.cpp
@@ -20,15 +20,27 @@
 
 #include "Mesh.h"
 
-#include <Game.h>
 #include <stdexcept>
-#include <stdint.h>
+#include <array>
+
+#include "OpenGL.h"
 
 using namespace openblack::graphics;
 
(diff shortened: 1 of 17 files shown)


src/Graphics/Mesh.cpp
```diff
@@ -89,14 +89,23 @@ void Mesh::Draw()
 
 void Mesh::Draw(uint32_t count, uint32_t startIndex)
 {
+	static const std::array<GLenum, 5> topologies = {
```
handsomematt: appears to error on Windows:

```..\src\Graphics\Mesh.cpp(99): error C2676: binary '[': 'const std::array<GLenum,5>' does not define this operator or a conversion to a type acceptable to the predefined operator```

Also I'd imagine this could be a constexpr too.

handsomematt: My bad it actually errors on the bit below: `auto topology = topologies[static_cast<size_t>(_topology)];`

PR #98, opened 2019-09-10
Warnings

bwrsandman:
Fix most warnings on clang on linux.
Enable warnings on msvc.

### DIFF
--- a/components/ScriptLibrary/include/LHVM/VMScript.h
+++ b/components/ScriptLibrary/include/LHVM/VMScript.h
@@ -30,9 +30,17 @@ class VMScript {
 public:
 	VMScript() = delete;
 	VMScript(const std::string &name, const std::string &filename, uint32_t type, uint32_t var_offset, std::vector<std::string> variables,
-		uint32_t instruction_address, uint32_t parameter_count, uint32_t script_id) :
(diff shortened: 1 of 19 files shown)


bwrsandman:
Because gcc and clang produce different errors for unused parameters and have different whitelisting flags.
I didn't want to put different options for different compilers (except for msvc).
We can add it later though.
As for not doing unused-xyz warnings: It makes a lot of sense for a reverse engineered game to have unused variables during early stages.

bwrsandman:
I also didn't add `/WX` on Windows because I'm on Linux.

janisozaur:
Unknown `-Wno-*` options are silently ignored, at least until you hit other error, then they get displayed as well. You could provide all required forms and enable `-Werror`

bwrsandman:
They were not silently ignored when I tried them earlier. Maybe I made a typo.

janisozaur:
Maybe you hit the other case: there was an error caused by enabling another option, then unrecognized flags would get reported as error as well.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

> no diagnostic is produced for `-Wno-unknown-warning` unless other diagnostics are being produced. This allows the use of new `-Wno-` options with old compilers, but if something goes wrong, the compiler warns that an unrecognized option is present.

bwrsandman:
That's probably it

PR #102, opened 2019-09-13
Shader Cleanup

bwrsandman:
This change add consitency to shader code, mainly common names and prefixes:

The updating of uniforms (with the exception of `u_model`) have been moved to their own function
`Renderer::UploadUniforms` which gets called between passes. This reduces driver overhead which can cause the rendering to stall for uniform updates. Doing all the updates before starting rendering allows rendering to run uninterrupted.

The shader manager now takes care of setting `u_viewProj` since it is expected to be the same for every shader in a same pass (i.e. Reflection Pass, Main render Pass).

The Mesh draw command now takes care of binding the proper program by using a `ShaderProgram` parameter which must be set by the caller.

The style changes in the shaders are:
* The prefix `u_` for uniforms
* The prefix `s_` for samplers
* The prefix `a_` for vertex attributes
* The prefix `v_` for varyings.
* The position is called `position` and not `pos`
* The uv element is called `texcoord0`
* The name for view projection matrix is `u_viewProj`
* The name for model matrix is `u_model`
* The name for generic diffuse textures is `s_diffuse`
* The name for generic bump maps is `s_bump`
* The final color is ~~`gl_FragColor` since we're only using 1 target.~~ is `o_color`.
* etc

### DIFF
--- a/assets/shaders/line.frag
+++ b/assets/shaders/line.frag
@@ -1,9 +1,8 @@
 #version 150 core
 
 in vec4 v_color;
-out vec4 outColor;
 
 void main()
 {
-    outColor = v_color;
+    gl_FragColor = v_color;
 }
--- a/assets/shaders/line.frag
+++ b/assets/shaders/line.frag
@@ -2,7 +2,9 @@
 
 in vec4 v_color;
 
+out vec4 o_color;
+
 void main()
 {
-    gl_FragColor = v_color;
+    o_color = v_color;
 }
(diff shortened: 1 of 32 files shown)


handsomematt:
gl_FragColor is deprecated in GLSL 1.3 (OpenGL 3.0)

handsomematt:
really nice on the rest of the changes, but if you could just revert the gl_FragColor stuff

bwrsandman:
Done

PR #111, opened 2019-09-24
fix copy of shader from assets to binary dir

NeroBurner:
shaders weren't copied. The GLOB expression searched in `openblack/src/assets` instead of `openblack/assets`


bwrsandman:
+1
Can we address how unconventional it is to put the main cmakelists in src?

NeroBurner:
yes, the structure could be made a bit more separated, but I think this could be done later when the project is more mature and less rapidly changing

Also added some fixes for the install target, can start openblack directly from the istalled bin dir

PR #114, opened 2019-09-26
Convert rendering to bgfx

bwrsandman:
This is a prototype about getting the rendering to using a more sensible rendering library than OpenGL. I chose bgfx because I had heard about it and have been trying it out. It's somewhat mature and is less bulky than Ogre 2 or open/vulkan scene graph. Using bgfx, when sticking to the tested cases, works like a charm, but any time you try to do something fancy, you run into bugs or learn that you cannot do it as you intend. Exploration into Ogre, vulkan scene graph [Google's Filament](https://github.com/google/filament), [urho3d](https://urho3d.github.io), etc or even making our own should be done before throwing ourselves into any api.

Bgfx, is a pretty low level rendering library but does bring a few things for free (https://bkaradzic.github.io/bgfx/overview.html).
* There are many supported backends: Metal, WebGL, OpenGL, Vulkan, DirectX etc.
* There are many debug modes, stats, error logging.
* Render passes can be grouped into views
* There is a multi-threaded mode.
* Simplification in regards to shaders
    * Standardization of model, view and projection matrices
    * `varying.sc` glue shader code for attributes and varyings
    * Single shader can be cross compiled to all apis

Using it requires bgfx to be installed. The shaders need to be compiled. They can be compiled into a header (though this is not done as of yet).

There were a few bugs in bgfx that I ran into:
* https://github.com/bkaradzic/bgfx/issues/1829
* https://github.com/bkaradzic/bgfx/pull/1892
* https://github.com/bkaradzic/bgfx/issues/1891
* https://github.com/JoshuaBrookover/bgfx.cmake/pull/62

Hunter doesn't have bgfx as package but it's a desired feature:
* https://github.com/ruslo/hunter/issues/372

I have a branch of bgfx which I used with some fixes:
* https://github.com/bwrsandman/bgfx/commits/materials

Upstream support of Vulkan is fairly new. I managed to get stuff drawn and textures loaded but there were problems with the formats, textures (especially arrays) and the performance isn't great.

Not done yet is making the debug draw cross be drawn on top of everything. It's hard to do since setting the depth test state can be set but the order of draw is sorted. There would be a way to fix this by changing the view.

Some work has to be done for using it as a submodule or library. Getting it working with hunter/cmake (I use https://github.com/JoshuaBrookover/bgfx.cmake). Testing it on windows with Direct3D.

There is also some clean-up to do and some architecture refactoring to do.

Something to note is that editing shaders won't work without a rebuild. This is typical with modern graphics API. If this is something we really want (including live shader reloading), then we have to integrate `shaderc` as a library (discussed in https://github.com/bkaradzic/bgfx/issues/405). On the other hand, something like what is done for the imgui shader can be done for all our shaders, compiling them into the binary so there isn't any issue with finding the shaders at runtime.

image


bwrsandman:
309.4 fedora rawhide passes.

handsomematt:
Runs on Ubuntu okay, some obvious issues:

* Transparency / blending isn't working (anything completely transparent is discarded in fragment shader still)
* Winding order on meshes seems to be weird and culling themselves inappropriately.

bwrsandman:
> * Transparency / blending isn't working (anything completely transparent is discarded in fragment shader still)

I'll take a look when I have some time. What's a transparent object I can look at for tests?

> * Winding order on meshes seems to be weird and culling themselves inappropriately.

I'm not too sure what to look for here... Can you be more specific, so I can repro?

handsomematt:
image

You can see in this image no blending on tree and the weird culling on buildings.

PR #116, opened 2019-09-27
cmake: Prefix all openblack options with OPENBLACK

bwrsandman:
Better namespace for options.
Allows cmake-gui to group all `OPENBLACK` options together.
Differentiate `HUNTER` options from `OPENBLACK` options.
Set `HUNTER_ENABLED` instead of creating an option.

Replace `OPENBLACK_USE_SUBMODULES` with `OPENBLACK_USE_SYSTEM_DEPS` which is more clear.


NeroBurner:
what's the advantage of just setting the `HUNTER_ENABLED` variable versus making it an option?

NeroBurner:
we could add even more options in the form of `OPENBLACK_USE_SUBMODULE_XXX` to have the option to disable a specific submodule (and provide it yourself)

handsomematt:
I'd imagine using the built in Hunter option makes more sense then adding a custom one that does the exact same thing.

bwrsandman:
`HUNTER_ENABLED` is declared as an option in `cmake/HunterGate.cmake`
Setting it here is the just putting a default option value for that.

PR #118, opened 2019-10-03
[WIP] Fix Travis

raffclar:


NeroBurner:
I think you want to update the docker images over there https://github.com/openblack/openblack-docker

raffclar:
@NeroBurner  Why do we have Dockerfiles in this repo?

NeroBurner:
fixing over there https://github.com/openblack/openblack/pull/119

In the beginning I wrote the dockerfiles and they were built on each CI build. This took too long and to reduce the amount of time we need to wait for a build to finish we build those images once a day and upload them to dockerhub. The work for this was mostly done by janisozaur (I think)

raffclar:
Ok fairs. Want me to fix this (thinking of using git-core/ppa for all)?

NeroBurner:
maybe we could keep EnTT as submodule? I don't like fetching Entt `master` branch directly. This could break from one build to another, even on the same commit (on our branch)

PR #120, opened 2019-10-04
Refactor CMake and dependencies; find_package prefer CONFIG (over Hunter)

handsomematt:
Come to the conclusion that dependency management is best left out of CMake and left up to the user, I added a small snippet to automatically use vcpkg if the environment variable is present.

The user can provide packages through:
* Installing them as system dependencies through their package manager: `apt install lib-sdl2`
* Installing them through vcpkg: `vcpkg install sdl2`
* Providing library directories themselves through CMake variables.

Hunter was restricting of available packages and generally versions behind, where as vcpkg has every package we need to depend on (except for bgfx at the moment but microsoft/vcpkg#7763 seems to be good)

Todo, I believe we need CMake modules for the following (to find system):
* [x] bgfx
* [x] sdl2
* [x] spdlog
* [x] entt
* [x] imgui


handsomematt:
I've set `CMAKE_FIND_PACKAGE_PREFER_CONFIG` which is a CMake 3.15 feature - this is so if using vcpkg it will search for config files first instead of cmake/ modules. This could fallback on the following however:

```# config packaged provided by zlib on github: https://github.com/madler/zlib
find_package( zlib CONFIG )
set( zlib_import_target zlib )

if( NOT zlib_DIR )
    # As a last resort, use the module package
    find_package( ZLIB REQUIRED )
    set( zlib_import_target ZLIB::ZLIB )
endif()
```

PR #126, opened 2019-10-20
Add convenience commandline parameters

bwrsandman:
`-v`: enable vsync (defaulted to false)
`-b`, `-f`: windowing borderless or fullscreen respecively. If both are supplied, borderless takes precedence. If none are supplied, windowed is used.

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -70,12 +70,17 @@ Game::Game(int argc, char** argv)
 
 	int windowWidth = 1280, windowHeight = 1024;
 	bool vsync = false;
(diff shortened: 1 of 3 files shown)


bwrsandman:
Can we agree to use single `-` for one char parameters but `--` for multi? The way the code is written now, help would have to be `-help`. I prefer `--help` being help and `-help` being equivalent to `-h -e -l -p`.

handsomematt:
sure sounds good, are there any good header only command line parsers out there that do it for us? (like boost program_options but not as heavy)

PR #131, opened 2019-10-24
L3DSubMesh: don't preserve state if texture changes

bwrsandman:
Fix s_diffuse already set error.


bwrsandman:
This should fix @raffclar error

src/3D/L3DSubMesh.cpp
```diff
@@ -161,23 +161,27 @@ void L3DSubMesh::Submit(uint8_t viewId, ShaderProgram& program, uint64_t state,
...
 				texture = &Game::instance()->GetMeshPack().GetTexture(prim.skinID);
 
-			program.SetTextureSampler("s_diffuse", 0, *texture);
+			if (lastTexture != texture)
+				program.SetTextureSampler("s_diffuse", 0, *texture);
 		}
 
 		_indexBuffer->Bind(prim.indicesCount, prim.indicesOffset);
-		bgfx::submit(viewId, program.GetRawHandle(), 0, preserveState || std::next(it) != _primitives.end());
+		bgfx::submit(viewId, program.GetRawHandle(), 0, lastTexture == texture && (preserveState || std::next(it) != _primitives.end()));
```
handsomematt: this line concerns me, the point of state preservation is to maintain the transforms and other state information, does this not screw with that?

bwrsandman: Texture changes qualify as state changes in bgfx it seems

bwrsandman:
Needs fix. The code needs to look at the next texture, not the previous.

PR #149, opened 2019-11-03
Mesh Reversing Utility

bwrsandman:
Change strange enum use in submesh to a struct with bitfields.
Add imgui widget to filter meshes based on their flags. I'm interested in finding a transparency flag that is mentioned in https://www.bwfiles.com/files/file.php?id=1266

image

Not in the image: Hovering over a bitfield button has a tooltip of the name.


handsomematt:
`static_assert(sizeof(HeaderFlag) == 4);` seems to be failing on MSVC Windows.

`sizeof(HeaderFlag)` is returning 5

bwrsandman:
Glad that assert is there then

PR #150, opened 2019-11-05
Registry context refactor

raffclar:


src/Entities/RegistryContext.h
```diff
@@ -0,0 +1,45 @@
+#include <Graphics/DebugLines.h>
+
+namespace openblack::Entities {
+class RenderContext
+{
+public:
```
bwrsandman: A class with everything public is a struct. You could use a struct here

src/Entities/RegistryContext.h
```diff
@@ -0,0 +1,45 @@
...
+	/// its allocated size. It will never shrink.
+	/// The values stored are a list of uniforms (model matrix) needed for both
+	/// the instances of entities and their bounding boxes.
+	bgfx::DynamicVertexBufferHandle instanceUniformBuffer;
+};
+
+using TownId = int;
+using StreamId = int;
+
+class RegistryContext
```
bwrsandman: and here

src/Entities/Registry.cpp
```diff
@@ -18,19 +18,15 @@ namespace openblack::Entities
 {
 
 Registry::Registry()
-	: _instanceUniformBuffer(BGFX_INVALID_HANDLE)
-	, _dirty(true)
+	: _dirty(true)
 	, _hasBoundingBoxes(false)
 {
 	_registry.set<RegistryContext>();
+	_registry.ctx<RegistryContext>().renderContext.instanceUniformBuffer = BGFX_INVALID_HANDLE;
```
bwrsandman: This should be done in the in the render context constructor

src/Entities/RegistryContext.h
```diff
@@ -0,0 +1,45 @@
...
+	/// Stores information for rendering which is prepared at \ref PrepareDraw.
+	std::map<MeshId, const InstancedDrawDesc> instancedDrawDescs;
+	/// Not an actual vertex buffer, but a dynamic general purpose buffer which
+	/// stores uniform data as a GPU-side copy of \ref _instanceUniforms and
+	/// which is populated in \ref PrepareDraw and consumed in \ref DrawModels.
+	/// This buffer will resize if the size of \ref _instanceUniforms exceeds
+	/// its allocated size. It will never shrink.
+	/// The values stored are a list of uniforms (model matrix) needed for both
+	/// the instances of entities and their bounding boxes.
+	bgfx::DynamicVertexBufferHandle instanceUniformBuffer;
```
bwrsandman: These need to be destroyed with bgfx when the class is destructed.
See https://github.com/openblack/openblack/pull/150/files#diff-b3ae7df02339c5d79de929ff2f8401ecL30-L31

Otherwise you get leaks when you shut down:
```
 BGFX Shutdown...
 BGFX render thread exit
 BGFX LEAK: DynamicVertexBufferHandle 1 (max: 4096)
 BGFX 	  0:    0
 BGFX LEAK: VertexBufferHandle 1 (max: 4096)
 BGFX 	  0: 1357 
 BGFX Shutdown complete.
```

src/Entities/RegistryContext.h
raffclar: I'm going to add a constructor and move _dirty and _hasBoundingBoxes as well. Don't see anything wrong with them being structs.

src/Entities/Registry.h
```diff
@@ -58,52 +52,34 @@ class Registry
 	decltype(auto) Create() { return _registry.create(); }
 	template <typename Component, typename... Args>
 	decltype(auto) Assign(entt::entity entity, [[maybe_unused]] Args&&... args) { return _registry.assign<Component>(entity, std::forward<Args>(args)...); }
+	decltype(auto) Context() { return _registry.ctx<RegistryContext>(); };
+	decltype(auto) Context() const { return _registry.ctx<RegistryContext>(); };
 	void Reset()
```
bwrsandman: This whole function implementation should be moved to the cpp file. Since you're refactoring, do you mind?

src/Entities/RegistryContext.h
raffclar: This is done when the registry is cleared

bwrsandman: It needs to be done while shutting down too. Otherwise you get those warnings.

raffclar: Sorted

src/Entities/Registry.h
raffclar: Because they have deduced return types, I'll need to rework them

raffclar: Done

src/Entities/RegistryContext.h
```diff
@@ -0,0 +1,55 @@
+#include <Graphics/DebugLines.h>
```
bwrsandman: Don't know how much @handsomematt cares, but I always add the copyright header to new files

bwrsandman:
Can't test now. Just make sure that there is no more leak message after cleanly exiting openblack.

PR #153, opened 2019-11-06
ci: Use system deps for spirv tools

bwrsandman:
This should significantly speed-up the bgfx build


bwrsandman:
I expect windows to fail due to some parts of the spirv-cross missing in cmake

PR #154, opened 2019-11-06
cxxopts for better argument parsing

### LINKED ISSUE #127: More robust command line parsing
Relating to https://github.com/openblack/openblack/pull/126#issuecomment-544281893

>Can we agree to use single `-` for one char parameters but `--` for multi? The way the code is written now, help would have to be `-help`. I prefer `--help` being help and `-help` being equivalent to `-h -e -l -p`.

Here are a few libraries. Note that it's pretty simple to just roll one out ourselves too.
* https://pubs.opengroup.org/onlinepubs/9699919799/functions/getopt.html
* https://github.com/cofyc/argparse
* https://github.com/hbristow/argparse
* https://github.com/p-ranav/argparse
* https://github.com/alex85k/wingetopt
* https://github.com/skandhurkat/Getopt-for-Visual-Studio
* https://github.com/skeeto/optparse
* https://github.com/bkaradzic/bx/blob/702b676ec28abc9c63942cc3d31c855adaffd4ce/include/bx/commandline.h
* https://github.com/septag/sx/blob/master/include/sx/cmdline.h
* https://github.com/jarro2783/cxxopts
* https://github.com/brofield/simpleopt
* Not boost please!

Many of these libraries haven't been updated in a while. This may be ok since a commandline parser doesn't exactly need a lot of maintenance.

bwrsandman:
Closes #127


bwrsandman:
Probably requires the linux dockers to have cxxopt installed

PR #165, opened 2019-11-13
Add diffuse texture to Water, mix with reflection buffer.

handsomematt:
image

Mixing diffuse + reflection doesn't give us an ideal result, we'll want to mess around with different blending methods.

### DIFF
--- a/assets/shaders/fs_water.sc
+++ b/assets/shaders/fs_water.sc
@@ -1,14 +1,15 @@
-$input v_texcoord0
+$input v_texcoord0, v_texcoord1
 
 #include <bgfx_shader.sh>
 
-SAMPLER2D(s_reflection, 0);
+SAMPLER2D(s_diffuse, 0);
+SAMPLER2D(s_reflection, 1);
 
 void main()
 {
+	vec3 diffuse_colour = texture2D(s_diffuse, vec2(v_texcoord0.x, v_texcoord0.y)).rgb;
+	vec3 reflect_colour = texture2DProj(s_reflection, v_texcoord1).rgb;
+	// vec3 water_color = vec3(0.35, 0.612, 0.643);
 
-	vec3 reflect_colour = texture2DProj(s_reflection, v_texcoord0).rgb;
-	vec3 water_color = vec3(0.35, 0.612, 0.643);
-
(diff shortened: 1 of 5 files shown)


assets/shaders/fs_water.sc
```diff
@@ -1,14 +1,15 @@
...
 
-SAMPLER2D(s_reflection, 0);
+SAMPLER2D(s_diffuse, 0);
+SAMPLER2D(s_reflection, 1);
 
 void main()
 {
+	vec3 diffuse_colour = texture2D(s_diffuse, vec2(v_texcoord0.x, v_texcoord0.y)).rgb;
+	vec3 reflect_colour = texture2DProj(s_reflection, v_texcoord1).rgb;
+	// vec3 water_color = vec3(0.35, 0.612, 0.643);
```
bwrsandman: We can remove this no?

src/3D/Water.cpp
```diff
@@ -38,6 +40,13 @@ Water::Water()
 {
 	_reflectionFrameBuffer =
 	    std::make_unique<FrameBuffer>("Reflection", 1024, 1024, graphics::Format::RGBA8, graphics::Format::Depth24Stencil8);
+
+	// water texture
+	auto const& waterTextureData = Game::instance()->GetFileSystem().ReadAll("Data/Textures/Sky.raw");
+
+	_texture = std::make_unique<Texture2D>("Water");
+	_texture->Create(256, 256, 1, Format::RGB8, Wrapping::ClampEdge, waterTextureData.data(), waterTextureData.size());
```
bwrsandman: Can we name these values for width and height instead of plugging the special 256 number everywhere?

src/3D/Water.cpp
```diff
@@ -86,6 +96,7 @@ void Water::DebugGUI()
 {
 	ImGui::Begin("Water Debug");
 	ImGui::Image(_reflectionFrameBuffer->GetColorAttachment().GetNativeHandle(), ImVec2(256, 256), ImVec2(0, 1), ImVec2(1, 0));
+	ImGui::Image(_texture->GetNativeHandle(), ImVec2(256, 256), ImVec2(0, 1), ImVec2(1, 0));
```
bwrsandman: and reference the width and height here

PR #167, opened 2019-11-14
Prevent out-of-bounds writes

raffclar:


bwrsandman:
Good catch. I went ahead and generalized your guard against out of bounds to more reads.

raffclar:
Havent got my laptop to test but looks ok.

I'll give it a thorough test this evening but merge anyway

PR #170, opened 2019-11-15
Fix using system installed spirv-tools

bwrsandman:


bwrsandman:
It's failing now because the vcpkg version of spirv-tools is 2019.3-dev-1 and we need at least 2019.3, ideally 2019.4

bwrsandman:
@handsomematt not sure how to get vcpkg to update

bwrsandman:
Might revisit this later, but for now it's not fixing a huge problem and it is out of date.

PR #171, opened 2019-11-15
[WIP] Load balance files (info.dat)

handsomematt:
Work in progress, drastically improve how we read info.dat so we can extend to read in even more of the data as well as making it extendable to possibly read the scripts and compile an Info.dat


bwrsandman:
You can use the packfile class for this from #225

PR #173, opened 2019-11-22
Use find_package(bgfx CONFIG) and make Azure Pipelines faster

handsomematt:
depend on bgfx's install target from: https://github.com/openblack/bgfx.cmake

this is now baked into the compile farm docker image for Linux builds, and for Windows builds precompiled binaries are pulled from here: https://github.com/openblack/bgfx.cmake/releases/tag/latest

our current version of bgfx is forked here: https://github.com/openblack/bgfx and includes some changes by @bwrsandman  https://github.com/bkaradzic/bgfx/compare/master...openblack:master

with no longer needing to build bgfx alongside openblack we have cut CI times down to 2 minutes:
image

couple of things before merging this:
* ~~need to add option `OPENBLACK_BUILD_BGFX` to build bgfx if people don't want to make bgfx for whatever reason~~ 
001d3a1
* ~~weird graphical bugs with latest bgfx~~ (specifically https://github.com/bkaradzic/bgfx/pull/1919 fixed by reverting and reapplying always bind)
* ~~need to rework `gpuTimeElapsed` as it's different in bgfx then our previous fork~~ 
d44a65c

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,19 +69,42 @@ else()
   endif()
 endif()
 
-find_package(bgfx CONFIG)
-if(NOT DEFINED bgfx_DIR)
-  # error if we're not building ourselves
-  if(NOT OPENBLACK_BUILD_BGFX)
-message(FATAL_ERROR " \
+option(OPENBLACK_BUILD_BGFX
+"Build bgfx alongside project (not advised)" OFF)
+
+if(NOT OPENBLACK_BUILD_BGFX)
+  find_package(bgfx CONFIG)
+  if(NOT bgfx_DIR)
+    message(FATAL_ERROR " \
 Failed to find bgfx, either install from: https://github.com/openblack/bgfx.cmake \
 Or set the variable OPENBLACK_BUILD_BGFX to build it with the project. \
 ")
   endif()
+else()
+  FetchContent_Declare(
+    bgfx
+    GIT_REPOSITORY https://github.com/openblack/bgfx.cmake.git
+    GIT_TAG master
+  )
 
-  # todo: reimplement building of bgfx here
-  # https://github.com/openblack/openblack/blob/2a363d58d213331781ea932c0c0cbb4a6a232994/CMakeLists.txt#L69
+  # don't build anymore targets then what we need
+  set(BGFX_BUILD_EXAMPLES
+      OFF
+      CACHE BOOL "Build bgfx examples.")
+  set(BGFX_INSTALL
+      OFF
+      CACHE BOOL "Create installation target.")
+  set(BGFX_CUSTOM_TARGETS
+      OFF
+      CACHE BOOL "Include convenience custom targets.")
 
+  # declare this here so bgfx compiles in a reasonable time
+  if(MSVC)
+    add_compile_options(/MP)
+  endif()
+
+  FetchContent_MakeAvailable(bgfx)
+  add_library(bgfx::bgfx ALIAS bgfx)
 endif()
 
 FetchContent_Declare(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,70 +69,19 @@ else()
   endif()
 endif()
 
-# todo: support find_package(bgfx)
-if(NOT bgfx_FOUND)
-  FetchContent_Declare(
-    bgfx
-    GIT_REPOSITORY https://github.com/openblack/bgfx.cmake.git
-    GIT_TAG materials
-  )
-
-  set(BGFX_BUILD_EXAMPLES
-      OFF
-      CACHE BOOL "Build bgfx examples.")
-  set(BGFX_INSTALL
-      OFF
-      CACHE BOOL "Create installation target.")
-  set(BGFX_CUSTOM_TARGETS
-      OFF
-      CACHE BOOL "Include convenience custom targets.")
-  set(BGFX_USE_DEBUG_SUFFIX
-      OFF
-      CACHE BOOL "Add 'd' suffix to debug output targets")
-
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -69,14 +69,31 @@ else()
...
+    message(STATUS "Precompiled bgfx binaries can be found here: https://github.com/openblack/bgfx.cmake/releases/tag/latest")
+  endif()
+
+    message(FATAL_ERROR " \
+Failed to find bgfx, either install from: https://github.com/openblack/bgfx.cmake \
+Or set the variable OPENBLACK_BUILD_BGFX to build it with the project. \
+")
+  endif()
+else()
+  message(AUTHOR_WARNING "Using OPENBLACK_BUILD_BGFX, if you are not developing use an install target instead.")
```
bwrsandman: Getting this message. Seems a little redundant since the dev is the one setting this variable
```
CMake Warning (dev) at CMakeLists.txt:88 (message):
  Using OPENBLACK_BUILD_BGFX, if you are not developing use an install target
  instead.
```

CMakeLists.txt
```diff
@@ -86,45 +103,6 @@ if(NOT bgfx_FOUND)
...
-    find_package(spirv_cross_reflect REQUIRED)
-    add_library(spirv-cross INTERFACE)
-    target_link_libraries(
-      spirv-cross INTERFACE spirv-cross-core spirv-cross-glsl spirv-cross-msl
-                            spirv-cross-reflect)
-    if(WIN32)
-      find_package(spirv_cross_hlsl REQUIRED)
-      target_link_libraries(spirv-cross INTERFACE spirv-cross-hlsl)
-    endif()
-  endif()
```
bwrsandman: Please keep this, it cuts down build time when you have spirv installed system wide

CMakeModules/Shaders.cmake
```diff
@@ -11,11 +11,9 @@
 
 # find_package(bgfx REQUIRED COMPONENTS shaderc)
 
-# get_target_property(BGFX_INCLUDE_PATH bgfx::bgfx INTERFACE_INCLUDE_DIRECTORIES)
-# set(BGFX_SHADER_INCLUDE_PATH ${BGFX_INCLUDE_PATH}/bgfx)
-
-add_executable(bgfx::shaderc ALIAS shaderc)
```
bwrsandman: This is needed if you are building bgfx from source

CMakeModules/Shaders.cmake
```diff
@@ -11,11 +11,9 @@
...
-set(BGFX_SHADER_INCLUDE_PATH ${bgfx_SOURCE_DIR}/bgfx/src)
+get_target_property(BGFX_INCLUDE_PATH bgfx::bgfx INTERFACE_INCLUDE_DIRECTORIES)
+list (GET BGFX_INCLUDE_PATH 0 BGFX_INCLUDE_PATH_1) # bgfx::bgfx exports include directory twice?
+set(BGFX_SHADER_INCLUDE_PATH ${BGFX_INCLUDE_PATH_1}/bgfx)
```
bwrsandman: Can you explain what `BGFX_INCLUDE_PATH_1` is? Why the numbered suffix?
Again, the include path is needed if debuggin bgfx source.

src/Gui.cpp
```diff
@@ -882,9 +882,11 @@ void Gui::ShowProfilerWindow(Game& game)
...
 			for (uint8_t i = 0; i < stats->numViews; ++i)
 			{
-				ImGui::Text("    %s: %0.3f", stats->viewStats[i].name,
-				            1000.0f * stats->viewStats[i].gpuTimeElapsed / (double)stats->gpuTimerFreq);
-				frameDuration -= stats->viewStats[i].gpuTimeElapsed;
+				auto const& viewStat = stats->viewStats[i];
+				int64_t gpuTimeElapsed = viewStat.gpuTimeEnd - viewStat.gpuTimeBegin;
+
+				ImGui::Text("    %s: %0.3f", viewStat.name, 1000.0f * gpuTimeElapsed / (double)stats->gpuTimerFreq);
+				frameDuration -= gpuTimeElapsed;
```
bwrsandman: :+1:

CMakeModules/Shaders.cmake
handsomematt: bgfx.cmake exposes `BGFX_INCLUDE_PATH` as a LIST, so it has two entries, when passed to shaderc it doesn't understand a list and just errors - so this simply gets the first entry

CMakeModules/Shaders.cmake
handsomematt: I'll move this to a custom BuildBgfx.cmake file

CMakeLists.txt
handsomematt: I'll move this to a custom BuildBgfx.cmake file - however I am slightly opposed to using this sort of thing at all as potentially your spirv-cross/tools could be different then the current version in bgfx's wonderful versionless repo

CMakeLists.txt
handsomematt: yeah just wanna make sure people aren't using it without a good reason

CMakeLists.txt
bwrsandman: fair, but I promise to keep the versions in sync ;)

CMakeModules/Shaders.cmake
bwrsandman: Why not export `BGFX_SHADER_INCLUDE_PATH` in bgfx.cmake That's the logical place to have it. openblack's use of the shader include files is not unique to its application. Every user of bgfx would need that include.

PR #179, opened 2020-02-03
Dynamics

bwrsandman:
Adding physics with Bullet.

I'm testing with bullet 3 but it should be compatible with bullet 2.

So far I have:
- [x] Land as static collision mesh
- [x] Ray casting the mouse to the land
- [x] Using surface normal to orient the debug cross

### DIFF
--- a/src/Dynamics/LandBlockBulletMeshInterface.h
+++ b/src/Dynamics/LandBlockBulletMeshInterface.h
@@ -1,0 +1,93 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2022 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include <cassert>
+
+#include <array>
+#include <vector>
+
+#ifdef _MSC_VER
+// Disable warning about conditional expression not being is constant
+#pragma warning(push)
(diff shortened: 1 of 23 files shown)


bwrsandman:
You can now gather information about objects by raycasting them.
I don't know entt enough to get more information about entities so that's left as an assignment.

Screenshot from 2020-05-20 17-50-34

bwrsandman:
Rebased to and tested current master.
Failing vcpkg linking on linux seems to be a vcpkg package bug. When linking system libs it works. 
Vcpkg issue happens when I test it too on latest vcpkg master.

raffclar (commented):
Compiled, ran, tested a few islands with no issues. Is it worth including info in the profiler?

src/Dynamics/LandBlockBulletMeshInterface.h
```diff
@@ -0,0 +1,83 @@
...
+			_indices[i] = i;
+		}
+	}
+
+	/// get read and write access to a subpart of a triangle mesh
+	/// this subpart has a continuous array of vertices and indices
+	/// in this way the mesh can be handled as chunks of memory with striding
+	/// very similar to OpenGL vertexarray support
+	/// make a call to unLockVertexBase when the read and write access is finished
+	void getLockedVertexIndexBase(unsigned char** vertexbase, int& numverts, PHY_ScalarType& type, int& stride,
```
raffclar, line 47: Lots of warnings about unused parameters

bwrsandman:
Definitely should include profiler

raffclar (approved):

PR #203, opened 2020-02-04
Ci format

bwrsandman:
Using https://github.com/microsoft/snmalloc/blob/master/azure-pipelines.yml as a reference

### DIFF
--- a/azure-pipelines-ci.yml
+++ b/azure-pipelines-ci.yml
@@ -4,6 +4,33 @@ pr:
 - master
 
 jobs:
+- job: format
+  displayName: 'Format'
+  pool:
+    vmImage: 'ubuntu-16.04'
+  container: 'openblack/compile-farm-ci:linux-amd64-gcc'
+
+  steps:
+    - task: CMake@1
+      displayName: 'CMake'
+      inputs:
+        cmakeArgs: '-DCMAKE_TOOLCHAIN_FILE=/tools/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-linux -Dbgfx_DIR=/bgfx-install/lib/cmake/bgfx ..'
+
+    - script: |
+        set -eo pipefail
+        cmake --build . --target clang-format
+        git diff -p --stat --color
+        git diff > format.patch
+        git diff --quiet $(Build.SourceVersion)
+      workingDirectory: build
+      failOnStderr: true
+      displayName: 'Clang-Format'
+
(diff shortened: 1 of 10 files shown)


azure-pipelines-ci.yml
```diff
@@ -55,3 +55,43 @@ jobs:
...
+
+  steps:
+    - script: |
+        set -eo pipefail
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+        sudo add-apt-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main"
+        sudo apt-get update
+        sudo apt-get install -y clang-format-8 clang-tidy-8
+        sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-8 100
+      displayName: 'Install clang'
```
bwrsandman: Would be better off if the docker had this installed already. @handsomematt

janisozaur: Angular has static builds of clang-format that get updated every so often: https://github.com/angular/clang-format

3-4MB binary is all you need. Keep in mind these are not directly tied to clang releases and _may_ produce slight variations in output in each version.

bwrsandman: Thanks. I'll try that. It does require npm though.

bwrsandman: Either @handsomematt has to install clang-format on the container or npm

janisozaur: The way I was using it was simply to download the binary directly from github, mark as a+x, then execute. No npm necessary really.

janisozaur: You can download whatever is on current master from this link: https://github.com/angular/clang-format/blob/master/bin/linux_x64/clang-format?raw=true

handsomematt:
I'll add clang-format to the docker image itself - in the mean time we need some sort of ignore list for third party headers such as `Common/stb_image_write.h` would probably end up moving all that to a src/ThirdParty folder too.

bwrsandman:
Good idea

handsomematt:
https://github.com/openblack/compile-farm/commit/e408b85d3ad7452b53295a2ebe3c7d27c6867a2c

bwrsandman:
This PR is mostly ready. I mentioned some changes to the formatting rules that I believe should be made (#208).

Before merging, the repo style should be fixed up (#207 and more) and the test run again to make sure nothing bad happens when publishing the diff with no errors.

bwrsandman:
Requires #215 merged to have green.

bwrsandman:
Rebased on the formatted code

PR #206, opened 2020-02-06
Instructions for Arch Linux using AUR package

bwrsandman:


handsomematt:
lovely, may want a seperate building.md at some point

PR #208, opened 2020-02-07
Clang format changes

bwrsandman:
While working on #203 which would enforce a style I noticed some settings in clang format which I wish to change.

These are my thoughts. Style is always a touchy subject but I will try to make my case. Take a look at the diff and I hope you will see what I see which is an improvement in readability in all cases.

I propose the following changes to clang format which all make code less horizontal and more readable:

* In constructors, put each initializer on a new line
This make initializers be vertical instead of all one one line.
It also makes diffs easier to read as a new initializer is one line on the diff which affects only that variable.

* Don't force short case statements on one line
Forcing short statements to have the form `case: code; break;` on the same line makes many switch statements harder to read and to modify. They don't even align properly and result in many cases to having to scroll horizontally.

* Don't force short blocks on one line
Similar to case statements, this makes many statements horizontal.
It makes some statements oddly aligned with a `{ ` at the beginning of a statement at the start of a new line.
If I use an `if` statement is one line and using braces it is because I would like it to be in a block,
not on one line.

### DIFF
--- a/src/Common/FileSystem.cpp
+++ b/src/Common/FileSystem.cpp
@@ -35,7 +35,10 @@ std::string FileSystem::FixPath(const std::string& path)
 		}
 	}
 
-	for (auto pos = result.find('\\'); pos != std::string::npos; pos = result.find('\\', pos + 1)) { result[pos] = '/'; }
+	for (auto pos = result.find('\\'); pos != std::string::npos; pos = result.find('\\', pos + 1))
+	{
+		result[pos] = '/';
+	}
 
 	return result;
 }
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -47,8 +47,11 @@ const std::string kWindowTitle = "openblack";
 Game* Game::sInstance = nullptr;
 
 Game::Game(Arguments&& args)
-    : _fileSystem(std::make_unique<FileSystem>()), _entityRegistry(std::make_unique<entities::Registry>()), _config(),
-      _frameCount(0), _intersection()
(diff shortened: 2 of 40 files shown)


src/Game.cpp
```diff
@@ -112,12 +116,19 @@ bool Game::ProcessEvents()
 		case SDL_KEYDOWN:
 			switch (e.key.keysym.sym)
 			{
-			case SDLK_ESCAPE: return true;
-			case SDLK_f: _window->SetFullscreen(true); break;
-			case SDLK_F1: _config.bgfxDebug = !_config.bgfxDebug; break;
```
bwrsandman: This is what I mean by unaligned code

src/Common/FileSystem.cpp
```diff
@@ -35,7 +35,10 @@ std::string FileSystem::FixPath(const std::string& path)
 		}
 	}
 
-	for (auto pos = result.find('\\'); pos != std::string::npos; pos = result.find('\\', pos + 1)) { result[pos] = '/'; }
```
bwrsandman: This is what I mean by horizontal code

src/Renderer.cpp
```diff
@@ -136,7 +139,9 @@ Renderer::~Renderer()
 void Renderer::LoadShaders()
 {
 	for (const auto& shader : Shaders)
-	{ _shaderManager->LoadShader(shader.name.data(), shader.vertexShaderName.data(), shader.fragmentShaderName.data()); }
```
bwrsandman: This is what I mean by a line starting by `{ `

PR #214, opened 2020-02-10
Add event queue and hand

raffclar:


src/Common/EventManager.h
```diff
@@ -0,0 +1,73 @@
...
+		inline static const int Type = identifier++;
+	};
+	mutable std::unordered_map<int, std::unique_ptr<IEventQueue>> queues {};
+	template <typename Event>
+	EventQueue<Event>* Insist()
+	{
+		const auto eventType = EventManager::_EventFamily::Type<Event>;
+		std::unique_ptr<IEventQueue> &p = queues[eventType];
+
+		if (!p)
```
raffclar: Nasty

src/Game.cpp
```diff
@@ -67,88 +71,97 @@ Game::Game(Arguments&& args)
 	}
 	_renderer = std::make_unique<Renderer>(_window.get(), args.rendererType, args.vsync);
 	_fileSystem->SetGamePath(GetGamePath());
+	_handModel = std::make_unique<L3DMesh>();
+	_handModel->LoadFromFile("Data/CreatureMesh/Hand_Boned_Base2.l3d");
```
bwrsandman, line 75: There's a question we have to answer here.
Should the test model also be drawn through the entity registry?
The sky?
The Land?

handsomematt: absolutely - all should probably have a MeshRenderer class that provides the various parameters required for bgfx submit / state

PR #220, opened 2020-02-13
CI: Rebase PR branch on target branch for pull requests.

handsomematt:


bwrsandman:
Interesting. Does azure not attempt to merge?

handsomematt:
nah, it'll just try to build your branch as is - figured it would be better if we rebase on the target branch since it'll tell us if it's safe to merge.

PR #222, opened 2020-02-13
logging: Add log to file and level options

bwrsandman:
By default openblack will log to a file.
Specifying `--logfile stdout` will log to terminal.
Default level is now debug and any number higher than 0 will be a higher
level such as info, warning and error.


bwrsandman:
This would be useful for testers who could send us logs.

handsomematt:
Would greatly prefer stdout as default for now, at the very least for debug builds.

bwrsandman:
That's a great idea. I'll add that in a bit.

PR #225, opened 2020-05-06
Animation

bwrsandman:
Reversed a good portion of the anm file format.
Added support for `AllAnims.anm` to `g3dtool` which we should rename to `packtool` and make a list of all files this file can open (`*.exc`, `*.cbn`, `*.hbn`, `*.hd`)
Added a way to view the animations on meshes in meshviewer, this will play any animation on any mesh if there are the same amount of bones.
There is time interpolation, but the time scale (frequency) might not be correct.
Might need to add an `AllAnims.anm` mock file before merging.

There is also a question of what the frame time value means. They don't always sum up to the animation duration, there is therefore sometime where the frame index to use is not clear.

anim1

anim0


bwrsandman:
Matrix bug should now be fixed

PR #226, opened 2020-05-08
Implement font rendering from custom game fonts.

handsomematt:
image

The game implements it's own custom font format, we read this in as glyphs and convert it to a texture atlas which we can then render directly on the screen.

I believe the game does box filtering on this font for anti aliasing too, however this isn't something I've implemented yet.

### DIFF
--- a/src/Util/Hash.h
+++ b/src/Util/Hash.h
@@ -1,0 +1,25 @@
+/******************************************************************************
+ * Copyright (c) 2018-2023 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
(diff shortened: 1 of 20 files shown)


src/Util/Hash.h
```diff
@@ -0,0 +1,24 @@
...
+#include <functional>
+
+namespace openblack
+{
+// https://www.boost.org/doc/libs/1_70_0/doc/html/hash/reference.html#boost.hash_combine
+template <class T>
+inline void hash_combine(std::size_t& seed, const T& v)
+{
+	std::hash<T> hasher;
+	seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
bwrsandman, line 22: Why not use `entt::cache` for this? https://github.com/skypjack/entt/wiki/Crash-Course:-resource-management

PR #238, opened 2020-05-20
Pathing

bwrsandman:


bwrsandman:
Closing because navmesh is not used in vanilla.
It uses a modified A* based on pre-generated graphs called footpaths.
The footpaths are made by joining town centre to building enterances

PR #241, opened 2020-07-08
Add morph component and tool

bwrsandman:
Morph files are the files found in `Data/CTR` which have animation, and hair data for the hand and creatures the file also lists up to 7 mesh variant (base, evil, good, thin, fat, weak, strong) file names (without extension) which are then found in `Data/CreatureMesh`.

The file specifies a spec file version which must be parsed after reading the version, this file is in `Data/` and is `ctrspec{version}.txt` for creatures and `hndspec{version}.txt`. The spec files contains information about the animations found in the rest of the morph file including categories, types and names.

Differentiating between a creature and a hand from the internal segment is done with a questionable field in the header, but could also be done by using the segment's name (from the pack library).

Hair data is largely unknown at this point. The structure sizes are known but their function are not.

Extra data is also unknown but they have some relation to the animations of the base animation set.

Next steps include using the morph library in the engine to load the hand mesh and creatures instead of using hardcoded paths to the .l3d.

The morph tool is not meant to read `.hbn` and `.cbn` files directly, instead it reads the contained segment.

More precisely, `.hbn` and `.cbn` files are pack files with single blocks. You can extract those blocks using packtool and that block is then read by morphtool. The reason for this design decision is to avoid coupling between pack and morph. In the future, multiple morph segments could be packed in a single pack file.

It's tedious to extract the segments from all the `cbn` files, so the pack tool and the morph tool now support piping. Here's a way to pipe the segment of a `cbn` file into morphtool using packtool:
```bash
$ packtool --extract stdout --block Creature Data/CTR/bmandrill.CBN | morphtool read --list-details stdin --spec-files-directory Data/

file: Data/CTR/bmandrill.CBN
name "Creature", size 2988308


Block writen to stdout
got 2988308 bytes from stdin
file: buffer
base mesh: "c_mandrill_boned_base"
6 variant meshes: "c_mandrill_boned_evil" "c_mandrill_boned_good" "c_mandrill_boned_thin" "c_mandrill_boned_fat" "c_mandrill_boned_weak" "c_mandrill_boned_strong"
232 animations from 21 categories:
4 variant animation sets
180 base animations
114 animations for "c_mandrill_boned_evil"
60 animations for "c_mandrill_boned_good"
4 animations for "c_mandrill_boned_thin"
29 animations for "c_mandrill_boned_fat"
3 hair groups
221 total extra data segments
232 extra data groups

```

Last note: I didn't start the write portion of the component or tool, this can be done later when integrating the component into openblack as it will be needed to create mock data for the Azure Pipeline.


raffclar:
Is this good for review?

bwrsandman:
Yes

raffclar:
Ok

raffclar:
This is really good

raffclar (approved):

PR #248, opened 2020-10-17
ci: add clang compiler test

bwrsandman:
This requires clang to be installed on the linux machine.
It will help catch clang only errors and shouldn't increase build times too much since windows build times take significantly longer than linux builds.

### DIFF
--- a/azure-pipelines-ci.yml
+++ b/azure-pipelines-ci.yml
@@ -82,12 +82,26 @@ jobs:
 
   strategy:
     matrix:
+      linux-amd64-clang-Debug:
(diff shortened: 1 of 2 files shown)


raffclar:
I have tried a few times to get this to work on Windows and I'm not having much luck because it's apparently too cutting edge to use clang-tidy with cmake integration.

Microsoft provide their own clang-tidy integration with visual studio. I intend on using it. Can we do a conditional on this to skip if on windows?

Will CI flag any clang-tidy issues?

bwrsandman:
It should skip those tests if clang-tidy is not found

handsomematt:
clang is now in the GitHub actions pipeline

PR #252, opened 2020-12-17
info.dat: Fully read file

bwrsandman:
Add all Infos structures
Move job of reading Infos to own class.
Fix MagicInfo size by adding 4 bytes.
Re-enable magic type debugging.

This is needed for work being done on villager states in which assigning villagers to abodes needs values from `GAbodeInfo::maxCapacity`.

Relates to #171

### DIFF
--- a/apps/packtool/packtool.cpp
+++ b/apps/packtool/packtool.cpp
@@ -285,6 +285,48 @@ int ViewAnimation(openblack::pack::PackFile& pack, uint32_t index, const std::st
 	return EXIT_SUCCESS;
 }
 
+int WriteRaw(const std::string& outFilename, const std::vector<std::string>& inFilenames)
+{
+	openblack::pack::PackFile pack;
+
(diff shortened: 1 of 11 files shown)


raffclar:
Ready for review?

bwrsandman:
I think @handsomematt wanted to give it a try first. Wait for his approval

raffclar:
> 
> 
> I think @handsomematt wanted to give it a try first. Wait for his approval

Ok

handsomematt (approved):
Good and important to get in, however not sure if we should mimic exactly how they did the serialization of data in global singleton classes or not, wonder if there are specific reasons they decided to go that route? Won't stop the merge, but worth considering keeping this flexible.

Also I feel like PackFiles are essentially LHSegmentFiles, poorly named?

bwrsandman:
I think it's `LHReleaseFile`

PR #258, opened 2020-12-24
component: Move components to their own files

bwrsandman:
Move mass of components out of AnimatedStatic.
Put Info in component.
Stringify enum names.


handsomematt:
Not sure about the enums being renamed, obviously they're better code wise, but there is then a mismatch with what Lionhead named them as, not sure if they are referenced anywhere. Info.dat?

bwrsandman:
Did I rename the enum definitions? I tried to only rename the enum types.

handsomematt:
> Did I rename the enum definitions? I tried to only rename the enum types.

Definitions were removed from the Enums.h file, these are 1:1 defined the same way as Lionhead would've defined them so I would rather keep that for now at least.

PR #262, opened 2020-12-27
Android

bwrsandman:
Depends on:
- [x] #358 ([diff](https://github.com/bwrsandman/openblack/compare/cross-compile-base...bwrsandman:android))

TODO:
- [x] vcpkg is blocking android for no good reason: https://github.com/microsoft/vcpkg/pull/23673
- [x] Runner requires newer ndk that has std::filesystem (`ANDROID_NDK_LATEST_HOME`) 
- [x] Some SDL fails due to the new SDL2 version might be fixed by installing the required libs: https://github.com/libsdl-org/SDL/blob/main/docs/README-linux.md
- [x] GLES 2.0 shader compile error
- [x] https://github.com/bkaradzic/bgfx.cmake/pull/129
- [x] #359

Follow-up:
- #280
- GLES rendering is wrong so Vulkan is default
- Imgui text resolution
- Figuring out the resolution / orientation: Is it using the device orientation and should we lock to landscape
- Make a native activity (no java): https://medium.com/androiddevelopers/getting-started-with-c-and-android-native-activities-2213b402ffff
- Stereo Rendering for VR on the Quest

Screenshot_20220329-224603_OpenBlack_1

To build, use Android studio and open the android directory


bwrsandman:
Open Issues:
- [ ] armeabi-v7a: Does not build due to SDL include requirement missing
- [ ] GLES 2.0 shader compile error
- [ ] Vulkan crashes emulated x86 phone
- [ ] Azure build of apks

vcpkg.json
```diff
@@ -2,15 +2,29 @@
     "name": "openblack",
     "version-string": "0.0.1-dev",
     "dependencies": [
-        "sdl2",
+        {
+            "name": "sdl2",
+            "features": ["x11"],
+            "platform": "linux"
```
bwrsandman, line 8: Add a "$reason" field with bug link

vcpkg.json
```diff
@@ -2,15 +2,29 @@
...
             "name": "bgfx",
             "default-features": false,
-            "features": ["tools"]
+            "features": ["tools"],
+            "platform": "native"
+        },
+        {
+            "name": "bgfx",
+            "default-features": false,
+            "platform": "!native"
```
bwrsandman, line 27: Add "$reason" field here

android/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,4 @@
+<resources>
+    <string name="app_name">OpenBlack</string>
+</resources>
```
bwrsandman, line 3: @handsomematt how would you like the app name to appear in an android app? In regards to capitalization

bwrsandman:
Rebase on top of c++20 just to be 100% sure it won't break

PR #263, opened 2020-12-28
Fix Warnings on Builds

bwrsandman:
I took a minute and went through all the warnings I get on Linux.
Found a few soft bugs and potential undefined behaviour.
I set an option to treat warnings as errors and enabled them on azure for Linux only.
I haven't gone through the warnings on MSVC, they seem too numerous and I don't want to deal with `fopen_s`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 include(CheckIncludeFileCXX)
 check_include_file_cxx(filesystem HAS_FILESYSTEM)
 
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 include(CheckIncludeFileCXX)
 check_include_file_cxx(filesystem HAS_FILESYSTEM)
 
+option(OPENBLACK_WARNINGS_AS_ERRORS "Treat warnings as errors (disable easier prototyping, enable for CI)" OFF)
+
```
bwrsandman, line 35: My question to the reviewer is should this be on by default and the dev turns it off themselves or leave it off by default?

handsomematt: Leave it off by default definitely.

PR #264, opened 2020-12-29
Use bundled fmt/fmt.h from spdlog

NeroBurner:
In GameThingSerializer `fmt/fmt.h` can't be found as it is not a direct
dependency provided by CMake. Use the bundled `fmt.h` from spdlog
instead.


bwrsandman:
Same as https://github.com/openblack/openblack/pull/262/commits/0fccf25b57895c35d9c3125a248616a7d1fa6952

PR #265, opened 2020-12-29
Add Map Grid Acceleration Structure

bwrsandman:
This matches the acceleration structure used in vanilla to look-up locally close fixed and mobile subclasses.
Vanilla used a linked list for this, this implementation uses a set to avoid duplicates, order should not matter either.

An accelerated look-up is needed to find town congregation point. A different AS could be used such as a BVH (used in modern physics and raytracing), quadtree (more dynamic), kd-tree, etc.

Depends on:
- [x] #258 for component clean-up

### DIFF
--- a/src/3D/L3DMesh.cpp
+++ b/src/3D/L3DMesh.cpp
@@ -29,6 +29,7 @@ L3DMesh::L3DMesh(const std::string& debugName)
     : _flags(static_cast<l3d::L3DMeshFlags>(0))
     , _debugName(debugName)
     , _physicsMass(1.0f) // TODO(bwrsandman): Find somewhere in file a value
+    , _boundingBox {glm::vec3(FLT_MAX, FLT_MAX, FLT_MAX), glm::vec3(-FLT_MAX, -FLT_MAX, -FLT_MAX)}
 {
 }
 
@@ -91,6 +92,10 @@ void L3DMesh::Load(const l3d::L3DFile& l3d)
 			_physicsMesh.reset(physicsMesh);
 			// FIXME(bwrsandman): Some meshes have multiple physics meshes
 		}
+		const auto& bb = subMesh->GetBoundingBox();
+		_boundingBox.minima = glm::min(_boundingBox.minima, bb.minima);
+		_boundingBox.maxima = glm::max(_boundingBox.maxima, bb.maxima);
+
 		_subMeshes.emplace_back(std::move(subMesh));
 	}
 	// TODO(bwrsandman): if no physics mesh was found, make physics mesh the bounding box
--- a/src/ECS/Archetypes/AbodeArchetype.cpp
+++ b/src/ECS/Archetypes/AbodeArchetype.cpp
@@ -13,6 +13,7 @@
 #include <spdlog/spdlog.h>
 
 #include "ECS/Components/Abode.h"
+#include "ECS/Components/Fixed.h"
 #include "ECS/Components/Mesh.h"
 #include "ECS/Components/Transform.h"
 #include "ECS/Registry.h"
@@ -50,6 +51,7 @@ entt::entity AbodeArchetype::Create(uint32_t townId, const glm::vec3& position, 
 	const auto& info = Game::instance()->GetInfoConstants().abode[static_cast<size_t>(type)];
 
 	registry.Assign<Transform>(entity, position, glm::mat3(glm::eulerAngleY(-yAngleRadians)), glm::vec3(scale));
+	registry.Assign<Fixed>(entity);
 	registry.Assign<Abode>(entity, info.abodeNumber, townId, foodAmount, woodAmount);
 	registry.Assign<Mesh>(entity, info.meshId, static_cast<int8_t>(0), static_cast<int8_t>(0));
 
--- a/src/ECS/Archetypes/AbodeArchetype.cpp
+++ b/src/ECS/Archetypes/AbodeArchetype.cpp
@@ -12,6 +12,8 @@
 #include <glm/gtx/euler_angles.hpp>
 #include <spdlog/spdlog.h>
 
+#include "3D/L3DMesh.h"
+#include "3D/MeshPack.h"
 #include "ECS/Components/Abode.h"
 #include "ECS/Components/Fixed.h"
 #include "ECS/Components/Mesh.h"
@@ -19,6 +21,7 @@
 #include "ECS/Registry.h"
 #include "ECS/Systems/TownSystem.h"
 #include "Game.h"
+#include "Utils.h"
 
 using namespace openblack;
 using namespace openblack::ecs::archetypes;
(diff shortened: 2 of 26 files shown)


src/Entities/Map.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <glm/vec3.hpp>
+
+#include <Entities/Components/Fixed.h>
+#include <Entities/Components/Transform.h>
+#include <Entities/Registry.h>
+#include <Game.h>
+
+glm::u16vec2 openblack::entities::Map::GetGridCell(const glm::vec3& pos)
+{
+	const glm::u32vec2 coords = glm::xz(pos) * static_cast<float>(0x10000) / 10.0f;
```
bwrsandman, line 22: This should be factored out as a static const

bwrsandman (commented):
Map should probably be moved to ECS/Systems

raffclar:
What happens if a fixed or mobile is removed?

raffclar (changes requested):

bwrsandman:
Every frame the map is recreated, so if an entity is removed, it's only invalid for one frame

src/Entities/Map.h
```diff
@@ -0,0 +1,42 @@
...
+#include <entt/fwd.hpp>
+#include <glm/fwd.hpp>
+#include <glm/vec2.hpp>
+
+namespace openblack::entities
+{
+
+class Map
+{
+	static constexpr glm::u16vec2 _gridSize = {0x200, 0x200};
```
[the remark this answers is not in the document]
bwrsandman: Yes in vanilla at least.

src/ECS/Map.cpp
```diff
@@ -0,0 +1,122 @@
...
+		    AddFixed(transform.position, entity);
+	    });
+	registry.Each<const components::Mobile, const components::Transform>(
+	    [this](entt::entity entity, [[maybe_unused]] const components::Mobile& mobile, const components::Transform& transform) {
+		    AddMobile(transform.position, entity);
+	    });
+}
+
+// TODO: Should pos be a parameter since we get transform anyways?
+void Map::AddFixed(const glm::vec3& pos, entt::entity entity)
```
bwrsandman, line 93: Remove pos

src/ECS/Map.cpp
```diff
@@ -0,0 +1,122 @@
...
+			{
+				auto& cell = _fixedGrid[cellId.x + cellId.y * _gridSize.x];
+				cell.insert(entity);
+			}
+		}
+	}
+}
+
+void Map::AddMobile(const glm::vec3& pos, entt::entity mobile)
+{
```
bwrsandman, line 118: Get transform instead of pos. Mark as not finished

src/ECS/Archetypes/AbodeArchetype.cpp
```diff
@@ -49,9 +56,28 @@ entt::entity AbodeArchetype::Create(uint32_t townId, const glm::vec3& position,
...
+	    registry.Assign<Transform>(entity, position, glm::mat3(glm::eulerAngleY(-yAngleRadians)), glm::vec3(scale));
 	registry.Assign<Abode>(entity, info.abodeNumber, townId, foodAmount, woodAmount);
 	registry.Assign<Mesh>(entity, info.meshId, static_cast<int8_t>(0), static_cast<int8_t>(0));
 
+	// Create Fixed component with a 2d bounding circle
+	const auto& bb = game->GetMeshPack().GetMesh(info.meshId).GetBoundingBox();
+	const auto bbSize = glm::max(glm::vec2(1.0f, 1.0f), glm::xz(bb.size() * 0.5f * transform.scale /** transform.scale*/));
+	if (glm::compMax(bbSize) / glm::compMin(bbSize) > 1.4)
+	{
+		throw std::runtime_error("Create non-square bounding circle");
```
bwrsandman, line 69: "NORSE_ABODE_C" throws here

src/ECS/Archetypes/AbodeArchetype.cpp
```diff
@@ -49,9 +56,28 @@ entt::entity AbodeArchetype::Create(uint32_t townId, const glm::vec3& position,
...
 
-	registry.Assign<Transform>(entity, position, glm::mat3(glm::eulerAngleY(-yAngleRadians)), glm::vec3(scale));
+	const auto& transform =
+	    registry.Assign<Transform>(entity, position, glm::mat3(glm::eulerAngleY(-yAngleRadians)), glm::vec3(scale));
 	registry.Assign<Abode>(entity, info.abodeNumber, townId, foodAmount, woodAmount);
 	registry.Assign<Mesh>(entity, info.meshId, static_cast<int8_t>(0), static_cast<int8_t>(0));
 
+	// Create Fixed component with a 2d bounding circle
+	const auto& bb = game->GetMeshPack().GetMesh(info.meshId).GetBoundingBox();
+	const auto bbSize = glm::max(glm::vec2(1.0f, 1.0f), glm::xz(bb.size() * 0.5f * transform.scale /** transform.scale*/));
```
bwrsandman, line 66: Delete that comment

src/ECS/Archetypes/AbodeArchetype.cpp
bwrsandman: In this case, an ellipse should be stored using x, z and angle

src/ECS/Archetypes/AbodeArchetype.cpp
```diff
@@ -49,9 +56,28 @@ entt::entity AbodeArchetype::Create(uint32_t townId, const glm::vec3& position,
...
+	}
+	else
+	{
+		glm::mat4 modelMatrix = glm::mat4(1.0f);
+		modelMatrix = glm::translate(modelMatrix, transform.position);
+		modelMatrix *= glm::mat4(transform.rotation);
+		modelMatrix = glm::scale(modelMatrix, transform.scale);
+		const auto point = glm::xz(glm::vec4(bb.center(), 1.0f) * glm::transpose(modelMatrix));
+		const auto radius = glm::compMax(bbSize);
+		registry.Assign<Fixed>(entity, point, radius);
```
bwrsandman, line 79: Perhaps, a component for elliptic fixed collide (for pathfinding) and a different one for circles.

src/ECS/Archetypes/AbodeArchetype.cpp
bwrsandman: In this case, a circle is approximated by taking `radius = sqrt(x * x + z * z)` and storing object angle.
I don't think we should store object angle since that should stay in transform and with the box being approximated, the angle shouldn't matter.

src/3D/L3DMesh.cpp
```diff
@@ -91,7 +92,9 @@ void L3DMesh::Load(const l3d::L3DFile& l3d)
 			_physicsMesh.reset(physicsMesh);
 			// FIXME(bwrsandman): Some meshes have multiple physics meshes
 		}
-		_subMeshes.emplace_back(std::move(subMesh));
```
bwrsandman, line 94: Oops! This line should not be removed

PR #266, opened 2020-12-31
l3d: decode bounding box in header

bwrsandman:


handsomematt (approved):

bwrsandman:
@handsomematt changed min to centre, max to size

PR #267, opened 2020-12-31
components: Add Mesh Component and use for drawing

bwrsandman:
By adding a mesh component, access to mesh id and sub mesh id are much quicker (struct is aligned on 8 bytes) and easier to do with EnTT.

Getting mesh Ids is now done at entity creation time.

Rendering is done with one `Each` call. No more need to use look-up tables. This greatly simplifies the amount of code needed.

Changing villager roles will have to be accompanied by changes to the mesh component.

The debug bounding boxes are now properly oriented after a fix.

Screenshot from 2020-12-31 00-31-34

Open question: for some reason some of the bounding boxes use different submesh ids. Why is this?

Depends on:
- [x] #258 for component clean-up

### DIFF
--- a/src/Common/MeshLookup.h
+++ b/src/Common/MeshLookup.h
@@ -28,13 +28,13 @@ namespace openblack
 class MeshId
 {
 public:
-	using IdType = int;
+	using IdType = int16_t;
 	constexpr MeshId(MeshPackId id)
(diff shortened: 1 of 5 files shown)


bwrsandman:
@handsomematt proposed a better solution. Input?

PR #270, opened 2021-01-02
Add registry functions

bwrsandman:
* Add missing header include guard.
* Add `Has` operation to check if entity has a component before casting
* Add `ToEntity` to go from component to entity
* Add `As` to go from component to another
* Add const variants and multiple template support


raffclar:
Ready for merge?

raffclar (approved):

PR #273, opened 2021-01-07
Add audio playback

raffclar:
Adds audio playback for SFX, voice and music. Uses OpenAL-soft with FFmpeg.

## Remaining work before we can merge
- [x] Fix mono sound quality. Sound has this hollow quality to it.
- [x] Fix mono sound pitch. Pitch sounds slightly too high. Likely interpreting pitch field incorrectly from .sad files.
- [x] Add ability to export sounds to .wav format for debugging.
- [x] Fix resource leak in OpenAL when creating many emitters in quick succession.
- [x] Move out variable domain mapping function to some place where it can be reused.
- [x] Fix hard-crash when reinitializing the audio player during audio playback.
- [x] Reinitializing the audio player shouldn't set the volume to 0.
- [ ] Add ability to switch to mock interface implementations during runtime.
- [x] Add dedicated handling for music so that music volume control can be implemented.
- [x] Add stop, pause to emitters and the debug gui.
- [x] Rename the AudioLoader interface to AudioDecoder

And finally rebase into multiple commits for review


bwrsandman (changes requested):
Audio profiler should have a larger default size when opened the first time.
Crashes when I touch interact with the island.

These are the fixes I needed to get it compiled and running
https://github.com/openblack/openblack/commit/4e1bc553dd3195b58c22dab274fe3a13d3606f80

CMakeLists.txt
```diff
@@ -44,11 +44,17 @@ endif()
 if(CMAKE_VERSION VERSION_GREATER 3.15 OR CMAKE_VERSION VERSION_EQUAL 3.15)
   set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
 
+  find_package(OpenAL REQUIRED)
+  find_package(FFmpeg REQUIRED COMPONENTS AVCODEC AVFORMAT AVUTIL)
+  find_package(implot REQUIRED)
```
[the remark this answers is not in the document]
raffclar: I ended up not using implot as this PR is getting a little big. Still planning on using it in a later PR

src/Common/MemoryStats.cpp
```diff
@@ -0,0 +1,5 @@
+//
+// Created by h-laptop on 31/12/2020.
+//
```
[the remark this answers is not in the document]
raffclar: MemoryStats will be in a later PR

src/Common/MemoryStats.h
```diff
@@ -0,0 +1,73 @@
...
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#ifdef __linux__
+#include <sys/sysinfo.h>
```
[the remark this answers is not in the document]
raffclar: MemoryStats will be in a later PR

src/Entities/Components/Hand.cpp
```diff
@@ -10,12 +10,16 @@
...
 	// Data/CreatureMesh/Hand_Boned_Base2.l3d
 	// Data/CreatureMesh/Hand_Boned_Good2.l3d
 	// Data/CreatureMesh/Hand_Boned_Evil2.l3d
-
 	// load data\\hand.raw into a memory
 }
+
+const std::vector<audio::SoundId> Hand::_grabLandSounds = {"G_HandGrabLand_01.wav", "G_HandGrabLand_02.wav",
+                                                          "G_HandGrabLand_03.wav", "G_HandGrabLand_04.wav",
+                                                          "G_HandGrabLand_05.wav", "G_HandGrabLand_06.wav"};
```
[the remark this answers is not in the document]
raffclar: Oops done

src/Entities/Components/Hand.h
```diff
@@ -14,6 +14,7 @@
 #include "Graphics/Mesh.h"
 #include "Graphics/Texture2D.h"
 
+#include <Sound/Sound.h>
```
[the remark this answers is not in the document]
raffclar: Done

external/imgui/imgui.cpp
```diff
@@ -5865,7 +5865,7 @@ void ImGui::End()
         EndColumns();
     PopClipRect();   // Inner window clip rectangle
 
-    // Stop logging
+    // Destroy logging
```
raffclar, line 5868: Weird

src/GameWindow.cpp
```diff
@@ -48,6 +48,9 @@ GameWindow::GameWindow(const std::string& title, int width, int height, DisplayM
 
 		if (SDL_InitSubSystem(SDL_INIT_TIMER) != 0)
 			throw std::runtime_error("Could not initialize SDL Timer Subsystem: " + std::string(SDL_GetError()));
+
+		if (SDL_InitSubSystem(SDL_INIT_AUDIO) != 0)
+			throw std::runtime_error("Could not initialize SDL Audio Subsystem: " + std::string(SDL_GetError()));
```
raffclar, line 53: Don't need this (unless we want to try out sdl-mixer)

src/Sound/AudioDebug.h
```diff
@@ -0,0 +1,27 @@
...
+{
+public:
+	static void ShowDebugGui(Game& game);
+private:
+	static void AudioPlayer(Game& game, const std::vector<std::shared_ptr<SoundPack>>& soundPacks);
+	static void AudioSettings(Game& game);
+	static std::string _selectedPack;
+	static AudioEmitterId _selectedEmitter;
+};
+}
```
raffclar, line 27: Missing newline

src/Sound/FfmpegDecoder.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+	av_free(context->buffer);
+	avio_context_free(&context);
+}
+
+void FfmpegDecoder::SwrDeleteContext(SwrContext *context)
+{
+	swr_free(&context);
+}
+} // namespace openblack::audio
```
raffclar, line 160: missing newline

src/Sound/SoundPack.cpp
```diff
@@ -0,0 +1,112 @@
...
+	_isMusic = combinedData.size() > 0;
+
+	for (auto& [soundName, sound] : combinedData)
+		_sounds[soundName] = std::make_unique<Sound>(sound);
+
+	if (_sounds.empty())
+		spdlog::warn("Soundpack {} has no sounds", _name);
+}
+
+} // namespace openblack
```
raffclar, line 112: Update comment

src/Sound/SoundHandler.cpp
```diff
@@ -0,0 +1,199 @@
...
+	// Sort activated emitters
+	// TODO: Implement
+
+	// Release hardware resources
+	// Update emitters with hardware resources
+	// Assign hardware resources
+//	for (auto& [id, emitter] : _emitters)
+//	{
+//		_audioPlayer->CleanUpResource(emitter);
+//	}
```
raffclar, line 68: Re-work

raffclar:
@bwrsandman I've add FindFFmpeg.cmake. Can you test again

raffclar:
@bwrsandman Reworked the commits. Should help a bit

src/Sound/SoundHandler.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+const std::map<AudioEmitterId, AudioEmitter>& SoundHandler::GetEmitters() const
+{
+	return _emitters;
+}
+
+AudioEmitterId SoundHandler::CreateEmitters(const std::vector<SoundId> ids, glm::vec3& position, glm::vec3& velocity, glm::vec3& direction, glm::vec2& radius, float volume, bool world, PlayType playType)
+{
+	std::vector<SoundId> out;
+	std::sample(ids.begin(), ids.end(), std::back_inserter(out), 1, std::mt19937{std::random_device{}()});
```
bwrsandman, line 100: This random device should live in the class instead of always creating it.
Also, we should have a randomization service to centralize things.
This is also needed for AI.

---

Also, why not use this:
```c++
assert(ids.size > 0);
std::default_random_engine generator; // Need to seed and move to class or service
std::uniform_int_distribution<size_t> distribution(0, ids.size() - 1);
SoundId out = ids[distribution(generator)];
```
Which could obviously be abstracted within a random service to be something like
```c++
template<T>
const T& RandomService::PickOne(const std::vector<T> container, size_t minIdx = 0, size_t maxIdx = -1)
{
    assert(container.size() > 0);
    assert(minIdx < ids.size());
    if (maxIdx == -1)
    {
        maxIdx = ids.size() - 1;
    }
    assert(maxIdx < ids.size());
    std::uniform_int_distribution<size_t> distribution(minIdx, maxIdx);
    return container[distribution(m_generator)];
}
```

src/Sound/SoundHandler.cpp
```diff
@@ -0,0 +1,182 @@
...
+			iterator = _emitters.erase(iterator);
+		}
+		else
+			iterator++;
+	}
+}
+
+std::unique_ptr<Sound>& SoundHandler::GetSound(SoundId id)
+{
+	auto& sound = _soundIdLookup[id]->GetSound(id);
```
bwrsandman, line 81: I get a crash here during runtime because the map is accessed with `operator[]` without any checks for if id is within the map.

src/Sound/SoundPack.cpp
```diff
@@ -0,0 +1,112 @@
...
+		auto soundName = path.string();
+		auto ext = path.extension().string();
+		isSector = ext == ".mpg";
+		auto volume = sample.volume > 0 ? ((float)sample.volume) / 127.f : 1.f;
+
+		// We'll use the file name for the sound ID instead as each sector has different name
+		if (isSector)
+		{
+			if (combinedData.find(filename) == combinedData.end())
+				combinedData[filename] = {
```
bwrsandman, line 55: During my tests, at this point, `filename` is the whole path of the file. This is likely why fetching the id of the sound fails in other parts. You should be using the basename of the path.

---

Also if readability is a concern do this.
```c++
auto emplacement = combinedData.emplace(filename, Sound{});
auto& sound = emplacement.first->second;
if (emplacement.second) // Sound was not in the map
{
	sound.id = filename;
	sound.name = filename;
	sound.priority = sample.priority;
	sound.sampleRate = static_cast<int>(sample.sampleRate);
	sound.bitRate = 0;
	sound.volume = volume;
	sound.pitch = sample.pitch;
	sound.pitchDeviation = sample.pitchDeviation;
	sound.bytes = soundData;
	sound.channelLayout = ChannelLayout::Stereo;
	sound.playType = static_cast<PlayType>(sample.loopType);
	sound.loaded = false;
	sound.sectorMerged = true;
	sound.sectorCount = 1;
	sound.lengthInSeconds = -1;
}
else // Sound was already in the map
{
	sound.sectorCount++;
	sound.bytes.insert(sound.bytes.end(), soundData.begin(), soundData.end());
}
```
to replace the whole if statement `if (combinedData.find(filename) == combinedData.end()) ... else`

src/Sound/SoundPack.cpp
```diff
@@ -0,0 +1,112 @@
...
+				/*pitch=*/sample.pitch,
+				/*pitchDeviation=*/sample.pitchDeviation,
+				/*bytes=*/soundData,
+				/*format=*/ChannelLayout::Mono,
+				/*loopType=*/static_cast<PlayType>(sample.loopType),
+				/*loaded=*/false,
+				/*sectorMerged=*/false,
+				/*sectorCount=*/0,
+				/*lengthInSeconds=*/-1
+			};
```
bwrsandman, line 97: Here you can just do
```c++
Sound sound;
sound.id = soundName;
//...
```

bwrsandman (changes requested):
I get these errors when initializing
```
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[2021-01-17 10:58:58.707] [error] OpenAL error: AL_INVALID_OPERATION: the requested operation is not valid with call "alListener3f(AL_POSITION, pos.z, pos.y, pos.x)" at file "src/Sound/OpenAlPlayer.cpp" on line 221
[2021-01-17 10:58:58.707] [error] OpenAL error: AL_INVALID_OPERATION: the requested operation is not valid with call "alListener3f(AL_VELOCITY, vel.z, vel.y, vel.x)" at file "src/Sound/OpenAlPlayer.cpp" on line 222
[2021-01-17 10:58:58.707] [error] OpenAL error: AL_INVALID_OPERATION: the requested operation is not valid with call "alListenerfv(AL_ORIENTATION, listenerOri)" at file "src/Sound/OpenAlPlayer.cpp" on line 224
```

I still get a seg fault when clicking on the map. This is likely due to accessing a wrong key in the sound map due to the existing keys being full paths to the sound file when they should be basenames.

---

On a more general architecture point:
I think you should rework the emitter owership to have them be managed by the entity registry.
Your uses of the emitters can be the same but using the registry to loop and search through our ECS.
Emitter has velocity which can be a component and sounds like they need to update the `Transform` Component of the emitter.

src/Sound/SoundPack.cpp
bwrsandman: I prefer the suggested change with emplacement because already the comment for `sound.playType` was mislabled

CMakeLists.txt
```diff
@@ -84,8 +88,7 @@ Or set the variable OPENBLACK_BUILD_BGFX to build it with the project. \
   endif()
 else()
   message(AUTHOR_WARNING "Using OPENBLACK_BUILD_BGFX, if you are not developing use an install target instead.")
-
-  FetchContent_Declare(
+FetchContent_Declare(
```
bwrsandman, line 91: Unnecessary change

apps/sadtool/sadtool.cpp
```diff
@@ -0,0 +1,260 @@
...
+		("l,list-blocks", "List block statistics.")
+		("b,view-bytes", "View raw byte content of block.", cxxopts::value<std::string>())
+		("i,info-block", "List INFO block statistics.")
+		("B,body-block", "List Body block statistics.")
+		("M,meshes-block", "List MESHES block statistics.")
+		("m,mesh", "List mesh statistics.", cxxopts::value<uint32_t>())
+		("T,texture-block", "View texture block statistics.")
+		("t,texture", "View texture statistics.", cxxopts::value<std::string>())
+		("A,animation-block", "List animation block statistics.")
+		("a,animation", "List animation statistics.", cxxopts::value<uint32_t>())
```
bwrsandman, line 143: These options are copy pasted from g3d tool and don't make any sense for sadtool. You should look at the options for l3dtool or anm tool

src/Sound/OpenAlPlayer.cpp
```diff
@@ -0,0 +1,215 @@
...
+	alCheckCall(alGenSources(1, &emitter.audioSourceId));
+	alCheckCall(alGenBuffers(1, &emitter.audioBufferId));
+	alCheckCall(alBufferData(emitter.audioBufferId, soundFormat, soundData.data(), soundData.size(), sound.sampleRate));
+	alCheckCall(alSourcei(emitter.audioSourceId, AL_BUFFER, emitter.audioBufferId));
+	std::uniform_int_distribution<> dist(-sound.pitchDeviation, sound.pitchDeviation);
+	auto pitch = Math::MapTo((float)(sound.pitch + dist(_rand)) * 0.001, .0f, 1.0, .5f, 2.f);
+	alCheckCall(alSourcef(emitter.audioSourceId, AL_PITCH, pitch));
+
+	if (!emitter.world)
+		alCheckCall(alSourcei(emitter.audioSourceId, AL_SOURCE_RELATIVE, true));
```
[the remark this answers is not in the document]
raffclar: Good catch

src/Sound/SoundHandler.cpp
raffclar: I'm going to throw a custom std::runtime_error called `SoundNotFoundError` here. GetSound is called in a lot of different places and checks should be done during game startup to ensure we have the sounds.

src/Common/FileSystem.cpp
```diff
@@ -94,5 +94,20 @@ std::vector<std::byte> FileSystem::ReadAll(const fs::path& path)
 
 	return data;
 }
+std::vector<std::string> FileSystem::GetAllFilePaths(const std::string& path, const std::string& ext)
+{
+	std::vector<std::string> paths;
+
+	for (auto& entry : fs::recursive_directory_iterator(FindPath("Audio")))
```
bwrsandman, line 101: Can place all referenced to the "Audio" path with calls to
```
 static inline fs::path AudioPath() { return "Audio"; }
```

bwrsandman: Perhaps this function should return paths relative to Audio.

---

The `All` in the function name is confusing to me, it's only returning audio files.

src/Sound/SoundPack.cpp
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	if (!_sadFile.GetName().empty())
+		_name = _sadFile.GetName();
+	else
+		_name = (filename);
+
+	// TODO: Discover sector flag
+	bool isSector;
+	auto combinedData = std::map<std::string, Sound>();
```
bwrsandman: SadFile is duplicating what PackFile does.
You should be using PackFile to extract the segments and SadFile should support only reading these segments.
See how src/3D/AnimationPack.cpp does it

raffclar: SadFile has been removed. PackFile now does everything

src/Sound/AudioDebug.cpp
```diff
@@ -0,0 +1,287 @@
...
+SoundId AudioDebug::_selectedSound = -1;
+std::string AudioDebug::_selectedPack = "";
+AudioEmitterId AudioDebug::_selectedEmitter = InitialAudioEmitterId;
+const ImVec4 redColor = ImVec4(1.f, .0f, .0f, 1.f);
+const ImVec4 greenColor = ImVec4(0.f, 1.f, .0f, 1.f);
+
+void AudioDebug::AudioPlayer(Game& game, const std::vector<std::shared_ptr<SoundPack>>& soundPacks)
+{
+	auto& handler = game.GetSoundHandler();
+	if (ImGui::Button("Play") && _selectedSound != -1)
```
bwrsandman, line 27: error: comparison of integer expressions of different signedness: ‘openblack::audio::SoundId’ {aka ‘long unsigned int’} and ‘int’ [-Werror=sign-compare]

src/Sound/AudioDebug.cpp
```diff
@@ -0,0 +1,275 @@
...
+	ImGui::NextColumn();
+	ImGui::Text("World Location");
+	ImGui::NextColumn();
+	ImGui::Separator();
+	for (auto& [emitterId, emitter] : emitters)
+	{
+		if (ImGui::Selectable(("##" + std::to_string(emitterId)).c_str(), _selectedEmitter == emitterId, ImGuiSelectableFlags_SpanAllColumns))
+			_selectedEmitter = emitterId;
+		ImGui::SameLine();
+		ImGui::Text("%llu", emitterId);
```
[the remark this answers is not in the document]
bwrsandman: error: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘long unsigned int’ [-Werror=format=]

src/Sound/AudioDebug.cpp
```diff
@@ -0,0 +1,287 @@
...
+		                      ImGuiSelectableFlags_SpanAllColumns))
+			_selectedEmitter = emitterId;
+		ImGui::SameLine();
+		ImGui::Text("%u", emitter.audioSourceId);
+		ImGui::NextColumn();
+		ImGui::Text("%u", emitter.audioBufferId);
+		ImGui::NextColumn();
+		ImGui::Text("%s", handler.GetSound(emitter.soundId)->name.c_str());
+		ImGui::NextColumn();
+		ImGui::Text("%llu", handler.GetSound(emitter.soundId)->bytes.size());
```
bwrsandman, line 215: error: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘long unsigned int’ [-Werror=format=]

src/Sound/SoundHandler.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+#include <fstream>
+#include <vector>
+
+namespace openblack::audio
+{
+
+SoundHandler::SoundHandler(std::unique_ptr<AudioDecoder>&& decoder, std::unique_ptr<AudioPlayer>&& player)
+    : _audioDecoder(std::move(decoder))
+    , _audioPlayer(std::move(player))
```
bwrsandman, line 23: error: ‘openblack::audio::SoundHandler::_audioDecoder’ will be initialized after ‘openblack::audio::SoundHandler::_audioPlayer’

bwrsandman (changes requested):
There are a few errors that prevent compilation on my machine

components/pack/include/PackFile.h
```diff
@@ -43,6 +43,91 @@ struct G3DTextureHeader
...
+struct SadAudioMetaDataHeader
+{
+	char unknown[27];
+	uint32_t entries;
+};
+
+struct SadAudioBankSampleTable
+{
+	uint16_t numOfEntries;
+	uint16_t numOfEntries2;
```
bwrsandman, line 55: Is this value ever used?

Are we sure numOfEntries isn't just a single 32 bit int or maybe numOfEntries2 is just padding?

components/pack/include/PackFile.h
```diff
@@ -43,6 +43,91 @@ struct G3DTextureHeader
...
+	int16_t unknown6b;
+	uint32_t sampleRate;
+	int16_t unknownOthera;
+	int16_t unknownOtherb;
+	int16_t unknown7a;
+	int16_t unknown7b;
+	int32_t unknown8;
+	int32_t lStart;
+	int32_t lEnd;
+	char sampleDescription[255];
```
bwrsandman, line 97: this and `audioSampleName` are not aligned on 32 bits. Are we sure it's not a size 256? Either way most compilers will pad an extra char after the string

raffclar: Yep it needs to be 256

components/pack/include/PackFile.h
raffclar: It's not padding. Both fields are the same value for some reason.

src/Sound/FfmpegDecoder.h
```diff
@@ -0,0 +1,36 @@
...
+public:
+	std::string GetName() const override { return "FFmpeg Loader"; };
+	void ToPCM16(Sound& sound) override;
+
+private:
+	static void AvDeleteFormatContext(AVFormatContext* context);
+	static void AvDeleteAvioContext(AVIOContext* context);
+	static void SwrDeleteContext(SwrContext* context);
+};
+} // namespace openblack::audio
```
raffclar, line 36: Missing newline

src/Sound/AudioDebug.h
```diff
@@ -0,0 +1,30 @@
...
+
+private:
+	static void AudioPlayer(Game& game, const std::vector<std::shared_ptr<SoundPack>>& soundPacks);
+	static void AudioSettings(Game& game);
+	static PlayType _playType;
+	static SoundId _selectedSound;
+	static std::string _selectedPack;
+	static AudioEmitterId _selectedEmitter;
+};
+} // namespace openblack::audio
```
raffclar, line 30: Missing newline

raffclar:
@bwrsandman I tested it on Linux and sound playback was working. I'm refactoring the audio stuff into a component so I can write an app to play sounds without needing runblack.

bwrsandman:
I'll test when you're done that refactor. Give me a shout when.

components/sad/include/SADFile.h
```diff
@@ -0,0 +1,173 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

CMakeModules/FindFFmpeg.cmake
```diff
@@ -0,0 +1,118 @@
...
+    set(${_package_component}_DEFINITIONS  ${${_package_component}_PKGCONF_CFLAGS_OTHER})
+    set(${_package_component}_VERSION      ${${_package_component}_PKGCONF_VERSION})
+    libfind_process(${_package_component})
+endmacro()
+
+
+# setter for 'hashmap'
+macro(hashmap_set _table _key) # ARGN
+    set(${_table}_${_key} ${ARGN})
+endmacro()
```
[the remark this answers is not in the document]
bwrsandman: I don't agree with this

PR #277, opened 2021-01-30
sad: adjust use of pack and sad component and tool

bwrsandman:
Removed sound specific logic from packfile which is meant to only put many segment files together.
Moved that logic to down the line when reading the file.
```
* use packfile to unpack files into streams
* feed streams into sadfile to open using sad specific data
```


src/Sound/SoundPack.cpp
```diff
@@ -74,7 +92,7 @@ void SoundPack::LoadFromFile(const std::string& filePath)
 				sound.volume = volume;
 				sound.pitch = pitch;
 				sound.pitchDeviation = sample.pitchDeviation;
-				sound.bytes = soundData;
+				sound.bytes = std::vector<uint8_t>(soundData.begin(), soundData.end());
```
bwrsandman, line 95: Just so you know, this is a copy of the span into separate vectors. If you want to take advantage of spans you'll have to store the larger buffer as a copy and copy the spans

bwrsandman: I didn't do this here because it would be out of scope

PR #278, opened 2021-01-30
Add vcpkg manifest and submodule

### LINKED ISSUE #271: Add a vcpkg.json manifest
https://github.com/microsoft/vcpkg/blob/master/docs/specifications/manifests.md

handsomematt:
Adds a vcpkg manifest and vcpkg as a submodule, also makes CMake default the CMAKE_TOOLCHAIN_FILE to vcpkg in the submodule.

* Creates a way to easily maintain a list of dependencies.
* Creates consistent versioning across developers and CI/release environments.
* Easier for people to begin developing, they don't need to install vcpkg, CMake runs all the necessary install commands.

Resolves #271

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,10 @@ if(DEFINED ENV{VCPKG_ROOT} AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
   set(CMAKE_TOOLCHAIN_FILE
       "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
       CACHE STRING "")
+else()
+  # set default toolchain to vcpkg submodule
+  set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/scripts/buildsystems/vcpkg.cmake
+    CACHE STRING "Vcpkg toolchain file")
 endif()
 
 project(openblack)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,16 +7,18 @@ else()
   include(FetchContent)
 endif()
 
-# automatically use vcpkg if VCPKG_ROOT is defined
-if(DEFINED ENV{VCPKG_ROOT} AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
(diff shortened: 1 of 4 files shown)


bwrsandman:
Can you remove the default for us non vcpkg developers?

handsomematt:
> Can you remove the default for us non vcpkg developers?

Is there a better way to do it then to remove it completely? I want the project to be as easy as opening CMake and pointing it at the project with no configuration required - otherwise it's really intimidating for new developers.

vcpkg.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "openblack",
+    "version-string": "0.0.1-dev",
+    "dependencies": [
+        "sdl2",
+        "spdlog",
+        "glm",
+        "entt",
+        "cxxopts",
+        "openal-soft"
```
raffclar, line 10: ffmpeg?

handsomematt: ah, good catch - what specific parts of ffmpeg do we use as we can filter these.

CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ if(DEFINED ENV{VCPKG_ROOT} AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
   set(CMAKE_TOOLCHAIN_FILE
       "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
       CACHE STRING "")
+else()
+  # set default toolchain to vcpkg submodule
+  set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/scripts/buildsystems/vcpkg.cmake
+    CACHE STRING "Vcpkg toolchain file")
```
bwrsandman, line 19: I understand the desire to do this. I would make it easier for new devs, but it is essentially locking everyone into using vcpkg.
Things like `CMAKE_TOOLCHAIN_FILE` are and `CMAKE_CONFIGURATION_TYPES` should not be forced.
Isn't checking for vcpkg in the environment variables good enough?

bwrsandman (changes requested):
Not a fan of defaulting vcpkg, feels like locking in all users to vcpkg which makes it difficult for non-vcpkg devs and packagers.

PR #283, opened 2021-03-06
Footpath link

bwrsandman:
This is part of the villager state machine work that I am doing.

The footpath link is a look-up entry for associating buildings (`MultiMapFixed` and `PlannedMuliMapFixed`) to one or more footpaths.
This is used when a living wants to navigate to a building using a footpath.

The `FotFile` stores `FootpathLinkSave` which has a list of footpaths and position.
The footpaths in the list are duplicates of those found in the footpath list.
Eventually, the footpath links are traversed to associate themselves with loaded buildings (this is not done yet as MultiMapFixed is not yet added).

The solution here was to add equality operators to the `GameThingSerializer` structs and use `std::find` to get the references to the original footpaths.

### DIFF
--- a/src/Entities/Components/Footpath.h
+++ b/src/Entities/Components/Footpath.h
@@ -9,12 +9,14 @@
 
 #pragma once
 
-#include <glm/vec3.hpp>
 #include <vector>
+
+#include <glm/vec3.hpp>
 
 namespace openblack::entities::components
 {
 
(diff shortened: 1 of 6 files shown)


bwrsandman:
Note, I might add some `MultiMapFixed` commits to this later, if it's not yet merged

raffclar (approved):

PR #287, opened 2021-03-25
custom-levels - Add ability to load custom lands

raffclar:


src/3D/MeshPack.h
```diff
@@ -64,6 +64,7 @@ class MeshPack
 	void loadTextures(const std::map<std::string, pack::G3DTexture>& textures);
 	void loadMeshes(const std::vector<std::vector<uint8_t>>& meshes);
 
+	bool _enableUnknownMeshes = false;
```
bwrsandman, line 67: Could you move default values to the constructor to be more consistent?

src/Game.h
```diff
@@ -40,6 +41,7 @@ class FileSystem;
 class GameWindow;
 class Gui;
 class EventManager;
+class LevelLocator;
```
bwrsandman, line 44: Forward declare not necessary if you're including the header

src/LHScriptX/FeatureScriptCommands.cpp
```diff
@@ -317,6 +318,12 @@ inline glm::vec3 GetSize(int size)
 	return glm::vec3(size, size, size) * 0.001f;
 }
 
+void FeatureScriptCommands::SetATownInfluenceMultiplier(int32_t townId, float mult)
+{
+	SPDLOG_LOGGER_ERROR(spdlog::get("scripting"), "LHScriptX: {}:{}: Function {} not implemented.", __FILE__, __LINE__,
```
bwrsandman, line 323: Could you check if `SPDLOG_LOGGER_ERROR` already includes the file and line?

src/Level.cpp
```diff
@@ -0,0 +1,3 @@
+#include "Level.h"
+
+using namespace openblack;
```
bwrsandman, line 3: You can remove this file / doesn't include the copyright

src/LevelLocator.cpp
```diff
@@ -0,0 +1,22 @@
+#include "LevelLocator.h"
```
bwrsandman, line 1: Missing copyright

src/LevelLocator.h
```diff
@@ -0,0 +1,30 @@
...
+#include <vector>
+
+#include "Level.h"
+
+namespace openblack
+{
+class LevelLocator
+{
+public:
+	void LoadLevels(FileSystem& fs);
```
bwrsandman, line 22: can fs be const?

bwrsandman (changes requested):

src/Level.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include <string>
+#include <filesystem>
+
+namespace openblack
+{
+class Level
+{
+public:
+	Level(std::filesystem::path path) : _name(path.filename().string()), _scriptPath(path) {}
```
bwrsandman, line 20: You could move path instead of taking a copy

src/Level.h
```diff
@@ -0,0 +1,27 @@
...
+	const std::string& GetName() const { return _name; };
```
bwrsandman, line 21: This could be just `_scriptPath.filename().string()` instead of keeping a copy

bwrsandman (changes requested):

src/LevelLocator.h
raffclar: Yep and the new fs function

src/Level.h
raffclar: So path is an r-value reference?

src/3D/MeshPack.h
raffclar: Done. Updated the constructor

src/LevelLocator.cpp
raffclar: Oops thanks

src/Level.h
raffclar: Using std::move from the caller

src/Level.h
raffclar: Good idea

bwrsandman:
Is it normal that a bunch of these maps cause crashes? 
```
terminate called after throwing an instance of 'openblack::lhscriptx::LexerException'
  what():  unexpected character: �
```
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  File ./Data/Landscape/dance.led not found
```

Also, is there a way to sort these?
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  File ./Data/Landscape/dance.led not found
```

raffclar:
@bwrsandman Here's what we could do 
* All .txt files in the playgrounds folder are treated as level files and are added to menubar dropdown.
* Campaign levels 1-5 and tutorial island are manually specified and added to the menubar dropdown.

There's no other way that I can see right now that would work.

bwrsandman:
Is it possible to open the files to validate?

raffclar:
We could create a `bool Script::IsScriptFile(fs::path& p)` function and use `bool Script::isCommand(const std::string& identifier)` over all line items.

src/Level.h
```diff
@@ -0,0 +1,32 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

src/LevelLocator.cpp
```diff
@@ -0,0 +1,46 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

src/LevelLocator.h
```diff
@@ -0,0 +1,31 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

src/Gui/Gui.cpp
```diff
@@ -672,21 +673,20 @@ bool Gui::ShowMenu(Game& game)
...
-			constexpr std::array<std::pair<std::string_view, std::string_view>, 3> PlaygroundIslands = {
-			    std::pair {"Two Gods", "TwoGods.txt"},
-			    std::pair {"Three Gods", "ThreeGods.txt"},
-			    std::pair {"Four Gods", "FourGods.txt"},
-			};
+			auto& resources = Locator::resources::ref();
+			auto& levelsManager = resources.GetLevels();
+			auto campaigns = std::vector<entt::resource_handle<const Level>>();
+			auto playgrounds = std::vector<entt::resource_handle<const Level>>();
+			levelsManager.Each([&campaigns, &playgrounds](entt::id_type id, entt::resource_handle<const Level> level) {
```
bwrsandman, line 680: Warnings here

src/Level.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <filesystem>
+#include <string>
+
+namespace openblack
+{
+class Level
+{
+public:
+	Level(const std::string name, const std::filesystem::path path, bool isCampaign)
```
bwrsandman, line 20: Copies don't need to be const.
You should also `std::move` them
Unless they are meant to be references

src/Game.cpp
```diff
@@ -456,6 +457,42 @@ bool Game::Run()
...
+	levelManager.Load("campaign/Land1", "Land 1", scriptsPath / "Land1.txt", true);
+	levelManager.Load("campaign/Land2", "Land 2", scriptsPath / "Land2.txt", true);
+	levelManager.Load("campaign/Land3", "Land 3", scriptsPath / "Land3.txt", true);
+	levelManager.Load("campaign/Land4", "Land 4", scriptsPath / "Land4.txt", true);
+	levelManager.Load("campaign/Land5", "Land 5", scriptsPath / "Land5.txt", true);
+	// Load Playgrounds
+	levelManager.Load("playgrounds/LandT", "Tutorial Island", scriptsPath / "LandT.txt", false);
+	levelManager.Load("playgrounds/TwoGods", "Two Gods", scriptsPath / "Playgrounds" / "TwoGods.txt", false);
+	levelManager.Load("playgrounds/ThreeGods", "Three Gods", scriptsPath / "Playgrounds" / "ThreeGods.txt", false);
+	levelManager.Load("playgrounds/FourGods", "Four Gods", scriptsPath / "Playgrounds" / "FourGods.txt", false);
```
bwrsandman, line 471: Can you put a todo for #400

src/Gui/Gui.cpp
```diff
@@ -33,6 +33,7 @@
 #include <SDL2/SDL_syswm.h>
 #endif
 
+#include "Locator.h"
```
bwrsandman, line 36: Could you match the surrounding code and use `<>`

src/Gui/Gui.cpp
```diff
@@ -672,21 +673,20 @@ bool Gui::ShowMenu(Game& game)
...
-			    std::pair {"Land 4", "Land4.txt"},
-			    std::pair {"Land 5", "Land5.txt"},
-			    std::pair {"God's Playground", "LandT.txt"},
-			    // clang-format on
-			};
-			constexpr std::array<std::pair<std::string_view, std::string_view>, 3> PlaygroundIslands = {
-			    std::pair {"Two Gods", "TwoGods.txt"},
-			    std::pair {"Three Gods", "ThreeGods.txt"},
-			    std::pair {"Four Gods", "FourGods.txt"},
-			};
```
bwrsandman, line 689: Nice!

src/Game.cpp
```diff
@@ -456,6 +457,42 @@ bool Game::Run()
 		animationManager.Load(i, animations[i]);
 	}
 
+	// Load the campaign levels
+	const auto scriptsPath = _fileSystem->FindPath(_fileSystem->ScriptsPath());
+	levelManager.Load("campaign/Land1", "Land 1", scriptsPath / "Land1.txt", true);
+	levelManager.Load("campaign/Land2", "Land 2", scriptsPath / "Land2.txt", true);
+	levelManager.Load("campaign/Land3", "Land 3", scriptsPath / "Land3.txt", true);
+	levelManager.Load("campaign/Land4", "Land 4", scriptsPath / "Land4.txt", true);
+	levelManager.Load("campaign/Land5", "Land 5", scriptsPath / "Land5.txt", true);
```
bwrsandman, line 466: If campaign don't have names, the names here are kind of random. 
Perhaps the name can be set at load time allowing us to add and remove campaign levels dynamically.

bwrsandman (changes requested):
Could you rebase on master. There might be new fails use to #387.
You might need to copy `test/mock/Scripts/Land1.txt` into Lands 1 to 5 and the playgrounds

src/Gui/Gui.cpp
raffclar: Is this fixed? I'm not seeing it but I bet it's unused `id`

bwrsandman: You can see warnings in `Files Changed`

src/Game.cpp
raffclar: I could use the filename instead?

raffclar: The problem is that `Scripts` folder contains many .txt files that are not levels. if we iterate over the Scripts directory looking for .txt files and we add them to the menu, they'll be a lot of noise.

raffclar:
> Could you rebase on master. There might be new fails use to #387. You might need to copy `test/mock/Scripts/Land1.txt` into Lands 1 to 5 and the playgrounds

Yep :( It's failing because of this

src/Game.cpp
bwrsandman: All of the ones that aren't `InfoScript*.txt` are actually scene files though

src/Gui/Gui.cpp
```diff
@@ -698,15 +698,15 @@ bool Gui::ShowMenu(Game& game)
 			auto& filesystem = Game::instance()->GetFileSystem();
```
bwrsandman, line 698: New warnings because filesystem is no longer used

bwrsandman (changes requested):

bwrsandman (approved):

bwrsandman:
The fail will go away if you rebase

PR #288, opened 2021-04-07
New GitHub Actions CI pipeline, vcpkg manifest support, bundle bgfx.cmake as submodule

handsomematt:
Replaces Azure Pipelines CI with GitHub Actions, removes an additional barrier letting us do everything on GitHub.

Currently the supported pipelines are:

* Windows x64
* Windows x86
* Ubuntu (GCC)
* Ubuntu (Clang)
* MacOS (Experimental)

The Ubuntu pipelines are using vcpkg for dependency management, I'm open to having different pipelines use system dependencies but don't really have an interest in maintaining them when the same vcpkg.json can be defined and work on every system.

Based on vcpkg manifest PR #278 which lets us easily list and lock packages.

Other changes include:

* imgui added as a submodule and updated to 1.82 (from 1.73)
* bgfx.cmake added as a submodule tracked from openblack/bgfx.cmake, this is used when OPENBLACK_USE_BUNDLED_BGFX is on, which it is by default. let's us keep developers on same versions - can still use bgfx_DIR to target an installation target.
* entt updated to 3.7.0, API breaking change.

The CI pipeline will build vcpkg dependencies defined in vcpkg.json and caches them, when this file or the submodule commit id changes the cache is invalidated and rebuilt. The same applies for bgfx.cmake, the CI will build and cache it using `OPENBLACK_USE_BUNDLED_BGFX`.


raffclar:
Amazing work  👍

raffclar (approved):
I'm happy

bwrsandman:
This PR break on Arch Linux :(

PR #289, opened 2021-04-14
Fix EnTT 3.7.0 changes for action branch

raffclar:


src/Entities/Registry.h
```diff
@@ -57,7 +57,7 @@ class Registry
 	template <typename... Components>
 	decltype(auto) Has(entt::entity entity) const
```
bwrsandman, line 58: Would be nice to change the function name too. All of is much better.

bwrsandman (changes requested):

bwrsandman:
Will wait for the arch package to update to test this thoroughly

PR #291, opened 2021-04-30
Partial Revert of #288 which broke the arch package.

bwrsandman:
Reverting a few commits which cause the Arch package to fail.

The changes of #288 brought some unintended issues and breakages in the System-wide dependency distribution of openblack on Arch Linux. This affect both developer flow and automatic packaging via AUR:

1. New step introduced to get imgui: `git submodule update --init externals/imgui` after git clone (to avoid using recursive which brings in a lot of useless stuff) which is doable but not super simple with packaging tools. Not to mention cloning imgui takes a while
2. New required option: `-DOPENBLACK_USE_BUNDLED_BGFX=OFF`. If you're using clion, you have to edit as many CMakeCache.txt files as you have configurations, and it will break the first time you try. The use of bundled bgfx should be only used if you really need to debug bgfx. For regular I recommend using vcpkg or creating a package distro package of bgfx,cmake if you plan on packing it.
3. Glm just flat out doesn't work in cmake without the FindGlm
4. The install target openblack/src/../external/imgui/LICENSE.txt is broken

Let's work on ironing out the issues that remain.


bwrsandman:
I confirm this works with the Arch Linux package, just needs the addition of `-DOPENBLACK_USE_BUNDLED_BGFX=OFF`

bwrsandman:
Still need a way to get `glm::glm` to work on arch with the official version

bwrsandman:
I confirm that this fixes the Arch Linux AUR package and dev envs.

bwrsandman:
Closed in favor of #299

PR #292, opened 2021-05-01
Actions: Add System-deps with Arch Linux packaging

bwrsandman:
This should help early detect build issues not related to vcpkg

The new configuration uses the official Arch Linux dev docker image.
* It adds a non-root user to build the packages
* It installs official dependencies
* It then builds dependencies from the user community repository
* It then builds bgfx from the user community repository
* It then builds openblack with the PR SHA using the script from the user community repository

The build time can be greatly reduced by updates to the shared lib config of bgfx.cmake such as using system-wide glslang/spirv-cross

### DIFF
--- a/.github/workflows/ci-system-deps.yml
+++ b/.github/workflows/ci-system-deps.yml
@@ -1,0 +1,43 @@
+name: System Deps CI
+on: [pull_request]
+
+jobs:
+  arch-packaging-system-deps:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


bwrsandman:
System Deps CI / arch-packaging-system-deps will pass with #291

handsomematt (approved):

bwrsandman:
Closed in favor of #299

PR #293, opened 2021-05-01
Simple Input System

CyrilPaulus:
Basic input system to allow the game logic to react to a predefined sets of inputs instead of directly using the keyboard key codes.
This also enable the use of a keymap file to allows the user to change the keys.

What was changed 

* Added an input system that manage the keyboard and mouse inputs.
* Updated the camera to use this system.
* Updated the game logic to use this system (**_handGripping** replaced with this system).
* Added a new program argument to specify the keymap file. If no keymap is specified, the default keymap defined in the code is used.
* Added two keymap files, one for qwerty another one for azerty.

Planned features
* Allow to use mouse button in the keymap file.
* Allow to use multiple key to a specified input (ex SHIFT+A could rotate the camera on the left).
* Improve keymap file to use a standard format (json ?)

I already did the merge request so I could have feedback on the code quality and check if all the convention were respected.
I guess the pull request should only be merged once the planned are implemented and after the review process of course.

### DIFF
--- a/src/System/InputSystem.cpp
+++ b/src/System/InputSystem.cpp
@@ -1,0 +1,128 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#include <SDL.h>
+#include <fstream>
+#include "InputSystem.h"
+
+using namespace openblack;
+
+
+InputSystem::InputSystem()
+{
+	_mouseMouvements = glm::vec2(0, 0);
+
+	InitKeyType(InputType::CameraForward, "CameraForward", SDLK_w);
+	InitKeyType(InputType::CameraBackward, "CameraBackward", SDLK_s);
+	InitKeyType(InputType::CameraLeft, "CameraLeft", SDLK_a);
+	InitKeyType(InputType::CameraRight, "CameraRight", SDLK_d);
+	InitKeyType(InputType::CameraDown, "CameraDown", SDLK_LCTRL);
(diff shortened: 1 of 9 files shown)


UplinkCoder:
How about renaming it from InputSystem to just Input?

CyrilPaulus:
I can if it fits more the coding style of the project.
I choose InputSystem for the name because this class contains all the logic needed to handle the inputs. For me a class named "Input" would only store the data linked the the input and no logic. I always try to differentiate class that represent only data and class that represent logic or system with this convention.

But once again, if it fits more the coding style, I'll change it.

Also I put this class in the "System" folder because I couldn't find a existing folder that seemed to fit this class. Should I leave it there or move it in another folder ?

bwrsandman:
Can you expand a little bit on the km file format? Is it commonly used? I am not familiar with it

CyrilPaulus:
It's a simple text format I made up myself to be able to persist the mapping. The idea was to use the the km file as a temporary implementation until the engine would have a dedicated user setting service that would store the keymap in a more standard format. Ideally the InputService shouldn't parse itself the file. It should be using a SettingService to retrieve the configuration and the SettingService should be the one who would parse all the configuration files.

bwrsandman:
Looks like you need to build the clang-format target, include an stl header and fix a conflict

src/System/InputSystem.cpp
```diff
@@ -0,0 +1,128 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

src/System/InputSystem.h
```diff
@@ -0,0 +1,74 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

bwrsandman:
Closing due to inactivity. If you feel this has been unfairly closed, feel free to re-open.

PR #300, opened 2021-11-01
ECS Architecture clean-up

bwrsandman:
In this PR, I restructured how we do ECS.

1. Rename Entities to ECS to better represent the function. (Entities/Components makes no sense)
2. Improved Scripting by enabling marshalling of Enums instead of int for types
3. Added archetypes. Archetypes are the closest thing to vanilla classes, they represent types of entities which will have a certain subset of components. This is widely used in ECS design.
4. MeshIds now come from the game data. I removed Mesh lookups as a result and identified `MeshID` data in the info data.
5. For the nth time, I revised how angles and scaling is interpreted, this time by looking closely as disassembly of vanilla. The one thing I'm still not 100% sure of is whether to use negative angles or not.


handsomematt (approved):
Haven't read all of it, but skimmed and everything looks much cleaner 👍 ecs makes more sense and I'm very happy stuff is being loaded from info.dat now.

PR #301, opened 2021-11-12
Add clang-format action

bwrsandman:
https://github.com/marketplace/actions/clang-format-lint

This doesn't use the cmake clang-format, so it requires slightly different formatting (especially in the ignore file).
It is possible to have it automatically commit formatting fixes with https://github.com/marketplace/actions/add-commit
It does run much faster (~15s) than using the cmake clang-format because it doesn't need to install or configure anything .

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,2 +1,5 @@
-tiny_gltf.h
-stb_image_write.h
+apps/anmtool/tiny_gltf.h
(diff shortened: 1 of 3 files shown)


bwrsandman:
Testing out reviewdog which adds comments on the PR itself in the place of formatting errors
Screenshot from 2021-11-12 20-05-49

handsomematt (approved):

PR #302, opened 2021-11-14
Using CMake Presets for dev and Action

bwrsandman:
CMake added preset file support.

This allows us to specify different configurations we want for CMake.

I've tested this in VSCode on Windows (vcpkg) and Linux (vcpkg, system), Visual Studio on Windows and Clion on Linux.

I've also converted bgfx to a local vcpkg port so no more submodule for that.

I've also adapted the test matrix to use the vcpkg presets and using the `run-cmake` action which has added features such as diff annotation.

In all, I think it makes configuring the project for the first time and for longer use much more simple.


bwrsandman:
This new CI will add comments on PR diff where errors happen
image

handsomematt (approved):
As long as x64 windows is done in CI too, this looks great

bwrsandman:
Both windows-latest debug and release use the x64 triplet.
The x86-windows triplet is only being run with debug configuration

PR #307, opened 2021-11-16
Update the readme for cmake presets building

bwrsandman:
Since VSCode works the same on Windows and Linux, I split the OS-specific section into vcpkg/system and IDE categories.


bwrsandman:
@raffclar maybe you would want to add a section for using XCode

bwrsandman:
The visual studio 2022 instructions should be very similar to 2019 but I haven't tried it out yet.

handsomematt (approved):
Perfect, even easier to get started now 👍

PR #312, opened 2021-11-17
Fix last of vs warnings

bwrsandman:
These should be the last warnings that come out of building with visual studio 2022.

There are two more warnings from the flame-graph library which I can fix on that repo since I own it.
* Should be fixed now


raffclar:
Checking now

raffclar (approved):

PR #321, opened 2022-01-08
Pull additional data such as footprint, uv2, name...

raffclar:


handsomematt (approved):
gj on finally cracking this extra data

bwrsandman (changes requested):
Good job!
It's missing a few things to be fully polished though.
Could you add the following:
- [ ] Doc at the top of L3DFile.cpp
- [ ] Implement the write function too for the new fields.
- [ ] Add the footprint and uv2 data to the class.
- [ ] Test read and write in l3dtool (just printing names and vector sizes) + writes.

bwrsandman (approved):

raffclar:
I still have l3dtool write to finish. Do you think it's necessary?

bwrsandman:
Yes

apps/l3dtool/l3dtool.cpp
```diff
@@ -781,6 +814,16 @@ int ExtractFile(const Arguments::Extract& args)
 	memcpy(indexBuffer.data.data(), l3d.GetIndices().data(), sizeOfIndices);
 	gltf.buffers.push_back(indexBuffer);
 
+	// TODO(raffclar): tiny_gltf has not yet added support for serialising binary data. Using strings with base64
```
bwrsandman, line 817: For to-do in 3rd parties, we should open and issue in their repo and one in ours linking theirs.
Then this comment should be updated with a link to the tiny_gltf issue.

components/l3d/src/L3DFile.cpp
```diff
@@ -41,6 +49,15 @@
  *
  *  TODO: Investigate
  *
+ * ------------------------ start of uv2 block ---------------------------------
+ *
+ *  TODO: Investigate
```
bwrsandman, line 54: Uv should be vec2 format. What do you see when you print out d'une contents?

apps/l3dtool/l3dtool.cpp
```diff
@@ -781,6 +814,16 @@ int ExtractFile(const Arguments::Extract& args)
...
+	std::map<std::string, tinygltf::Value> extras;
+	extras["footprint"] =
+	    tinygltf::Value(tinygltf::base64_encode(l3d.GetFootprintData().data(), l3d.GetFootprintData().size()));
+	extras["uv2"] = tinygltf::Value(tinygltf::base64_encode(l3d.GetUv2Data().data(), l3d.GetUv2Data().size()));
```
bwrsandman, line 821: Uv clearly should not be in extras. They should probably be in a buffer with an accessor, view.
You'll need to check the format of the data but `type = "VEC2, componentType = GL_FLOAT` is very likely.
Attribute can be `TEXCOORD_1`

apps/l3dtool/l3dtool.cpp
```diff
@@ -781,6 +814,16 @@ int ExtractFile(const Arguments::Extract& args)
...
+	auto& name = l3d.GetNameData();
+	auto nameData = std::vector<unsigned char>(name.begin(), name.end());
+	extras["name"] = tinygltf::Value(tinygltf::base64_encode(nameData.data(), nameData.size()));
```
bwrsandman, line 824: Mesh has a property name that can be used. Wouldn't that be a better place?

bwrsandman (changes requested):
I have a few concerns about using the extra field for uv2 and name. They seem like the correct place for footprints until we analyze it but names and uv2 have more appropriate areas in my opinion.

It's also missing the `"write"` portion of the command. It should be simply the inverse of write so not too difficult to add.

apps/l3dtool/l3dtool.cpp
bwrsandman: Also, is name data à string? Why use binary encoding?

raffclar: I want to preserve the size of the name block. I noticed there'sa lot of unused space.

bwrsandman: Is the size constant with all files?

raffclar: I can't say for certain unless I check every file but so far they're the same. I've only checked a very small subset however

apps/l3dtool/l3dtool.cpp
raffclar: Agreed. I'll get on it

apps/l3dtool/l3dtool.cpp
bwrsandman: If it's always the same size then that's just an issue with their encoding. I think we can have string instead of base64

apps/l3dtool/l3dtool.cpp
raffclar: https://github.com/syoyo/tinygltf/issues/360

components/l3d/src/L3DFile.cpp
raffclar: None of the meshes we're loading currenty have uv2 data. I think it may be present in the compressed temple meshes. IMO it's worth leaving this until I tackle the next PR which is related to temples and camera paths

apps/l3dtool/l3dtool.cpp
raffclar: The UV2 buffer is optional and is not present in most meshes. I don't know what the UV2 data currently looks like

components/l3d/src/L3DFile.cpp
bwrsandman: Add an issue number then :)

apps/l3dtool/l3dtool.cpp
bwrsandman: Are you going to keep it in extras then? If so, add the same issue number as above

apps/l3dtool/l3dtool.cpp
```diff
@@ -911,9 +912,10 @@ int ExtractFile(const Arguments::Extract& args) noexcept
 	extras["footprint"] =
 	    tinygltf::Value(tinygltf::base64_encode(l3d.GetFootprintData().data(), l3d.GetFootprintData().size()));
```
bwrsandman, line 913: static_cast<uint32_t>(l3d.GetFootprintData().size())

apps/l3dtool/l3dtool.cpp
bwrsandman: static_cast<uint32_t>(l3d.GetUv2Data().size())

bwrsandman (approved):

PR #322, opened 2022-01-08
Add minimal support for creature isle meshes and textures

raffclar:


bwrsandman (changes requested):
Could you add to the PackFIle.cpp doc at the top of the the source file

raffclar:
> PackFIle
For the PackFile, there's nothing to document but the L3DFile has some minor changes in CI, noticeably to do with DXT5 textures being present. Want me to document that?

bwrsandman:
Yeah a little information about the format would be a useful place to answer questions about the types of textures supported and such.

raffclar:
> Yeah a little information about the format would be a useful place to answer questions about the types of textures supported and such.

Nevermind the key change is in the PackFile

components/pack/src/PackFile.cpp
```diff
@@ -282,10 +285,28 @@ void PackFile::ExtractTexturesFromBlock()
 		DdsHeader ddsHeader;
 		ddsStream.read(reinterpret_cast<char*>(&ddsHeader), sizeof(DdsHeader));
 
-		// TODO(bwrsandman) the extra sizeof(uint32_t) is unaccounted for
-		if (header.ddsSize - sizeof(ddsHeader) - sizeof(uint32_t) != ddsHeader.pitchOrLinearSize)
```
raffclar, line 286: @bwrsandman 
I'm still reading this example but I don't see any checks like that here: https://docs.microsoft.com/en-us/windows/uwp/gaming/complete-code-for-ddstextureloader
I've had to remove this check as the ddsSize field is larger than the pitchOrLinearSize for DXT5 textures.

PR #323, opened 2022-01-08
Skip submeshes that fail to load

raffclar:


bwrsandman (changes requested):
If the submesh fails, it should at least log a message. Perhaps in warning or err.

bwrsandman (approved):

PR #324, opened 2022-01-08
Add support for loading loose mesh files

raffclar:


src/Common/StringUtilities.h
```diff
@@ -0,0 +1,41 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2021 openblack developers
```
bwrsandman, line 2: 2022

src/3D/MeshPack.h
```diff
@@ -63,6 +64,7 @@ class MeshPack
 private:
 	void loadTextures(const std::map<std::string, pack::G3DTexture>& textures);
 	void loadMeshes(const std::vector<std::vector<uint8_t>>& meshes);
+	bool LoadMeshes(std::vector<std::filesystem::path> paths);
```
bwrsandman, line 67: Inconsistent capitalization here. Perhaps you can fix that.

src/Common/StringUtilities.h
```diff
@@ -0,0 +1,41 @@
...
+ *****************************************************************************/
+
+#pragma once
+
+#include <algorithm>
+#include <string>
+
+namespace openblack
+{
+inline bool EndsWith(std::string const& str, std::string const& ending)
```
bwrsandman, line 17: Any reason why these are all inline?

bwrsandman: Kinda unusual to have it reordered, please change to: `const std::string&`

bwrsandman: When copying code from a source, it's good practice to link it in a comment:
https://stackoverflow.com/questions/874134/find-out-if-string-ends-with-another-string-in-c

src/Common/StringUtilities.h
```diff
@@ -0,0 +1,41 @@
...
+{
+	if (ending.size() > str.size())
+	{
+		return false;
+	}
+
+	return std::equal(ending.begin(), ending.end(), str.begin());
+}
+
+inline void LowerCase(std::string& str)
```
bwrsandman, line 37: We already do this in here.
https://github.com/openblack/openblack/blob/33e0a1213173001167f2c4ace5afc07a40ea2dbb/src/Game.cpp#L466-L467

Could you use that formulation which is more C++ than C. And refactor it here.

bwrsandman: Also, the name sounds more like it should return a LowerCase version, not modify it. `ToLowerCase` would be better for inplace.

To be complete, it would be good to do the same for Upper.
Also a function for first letter capitalization which we do here: https://github.com/openblack/openblack/blob/e8e8431c0f813fdec2c089d609dab241f7c8ad0d/src/3D/Sky.cpp#L46

src/3D/MeshPack.cpp
```diff
@@ -34,20 +37,78 @@ MeshPack::MeshPack(bool enableUnknownMeshes)
...
+		auto meshId = MeshId(_meshes.size());
+		_meshes.emplace_back(std::move(mesh));
+		return std::make_pair(true, meshId);
+	}
+
+	return std::make_pair(false, MeshId::Dummy);
+	;
+}
+
+bool MeshPack::LoadMeshes(std::vector<std::filesystem::path> paths)
```
bwrsandman, line 103: You're accidentally copying the list.
```c++
const std::vector<std::filesystem::path>& paths
```

src/3D/MeshPack.cpp
```diff
@@ -34,20 +37,78 @@ MeshPack::MeshPack(bool enableUnknownMeshes)
...
-	loadTextures(pack.GetTextures());
-	loadMeshes(pack.GetMeshes());
+	return false;
+}
+
+std::pair<bool, MeshId> MeshPack::LoadLooseMeshFile(const fs::path& path)
+{
+	SPDLOG_LOGGER_DEBUG(spdlog::get("game"), "Loading Mesh Pack from file: {}", path.generic_string());
+	auto pathExt = path.extension().string();
+	LowerCase(pathExt);
```
bwrsandman, line 67: You're already making a copy of the string by using `auto`. Also, keeping the namespace makes the code more readable.
This is why I think it's better to just do this:
```c++
const auto pathExt = StringUtils::LowerCase(path.extension().string());
```

bwrsandman (changes requested):
MeshPack is not the right place to load loose files.

A MeshPack is the `.g3d` collection of meshes. Loose files are individual l3d files that are compressed.

A different interface should do this. If there is a lot of duplicated code, a base class can be shared.

raffclar:
I threw that string lib together haha. I'm open to ideas on doing it properly. Should we have both inplace and new copy funcs?

src/3D/MeshPack.cpp
raffclar: Wouldn't `namespace openblack::string` be better than `StringUtils`?

src/3D/MeshPack.cpp
raffclar: Ouch

src/Common/StringUtilities.h
raffclar: I completely forgout about this. My bad

src/3D/MeshPack.cpp
bwrsandman: Usually, you call it string if you're making your own implementation.

bwrsandman:
I think inplace would only be useful if we identify a bottleneck during runtime.
Since I don't expect these will run in a poop during the main game loop, or for very long inputs, a copy is fine.
it's easier to work with

src/3D/Sky.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "Sky.h"
 
+#include <Common/StringUtils.h>
```
raffclar, line 12: fix

raffclar:
> MeshPack is not the right place to load loose files.
> 
> A MeshPack is the `.g3d` collection of meshes. Loose files are individual l3d files that are compressed.
> 
> A different interface should do this. If there is a lot of duplicated code, a base class can be shared.

How would you build a MeshId to Mesh lookup? We need a storing all the meshes together so perhaps refactoring out `	MeshesVec _meshes` from MeshPack into a different class?

bwrsandman:
We need to decouple mesh_id and mesh resources. Mesh_id are an enum for meshes in the pack file.

raffclar:
@bwrsandman With #341 merged, the MeshPack no longer exists. No more weird workarounds with MeshIds anymore either.

This PR is now simplified with one purpose in mind: extend the L3DLoader to support zlib-compressed .zzz mesh files.

src/3D/Sky.cpp
bwrsandman: Remove that line

src/Common/StringUtils.h
```diff
@@ -0,0 +1,28 @@
...
+
+[[nodiscard]] bool BeginsWith(const std::string& str, const std::string& beginning);
+
+[[nodiscard]] std::string UpperCase(const std::string& str);
+
+[[nodiscard]] std::string LowerCase(const std::string& str);
+
+[[nodiscard]] std::string Capitalise(const std::string& str);
+
+} // namespace openblack::string_utils
```
bwrsandman, line 28: Add a line break

src/Game.cpp
```diff
@@ -390,6 +391,8 @@ bool Game::Run()
 	auto& textureManager = resources.GetTextures();
 	auto& animationManager = resources.GetAnimations();
 
+	const auto citadelOutsideMeshesPath = _fileSystem->FindPath(_fileSystem->CitadelPath() / "OutsideMeshes");
+	meshManager.Load("temple", "temple", citadelOutsideMeshesPath / "b_first_temple_l3d.zzz");
```
bwrsandman, line 395: You should give it the name of the file `b_first_temple_l3d`

src/Resources/Loaders.cpp
```diff
@@ -25,10 +31,37 @@ entt::resource_handle<L3DMesh> L3DLoader::load(const std::string& debugName, con
...
+	else if (pathExt == ".zzz")
+	{
+		auto stream = Game::instance()->GetFileSystem().Open(path, FileMode::Read);
+		uint32_t decompressedSize = 0;
+		stream->Read(reinterpret_cast<uint32_t*>(&decompressedSize), sizeof(decompressedSize));
+		auto buffer = std::vector<uint8_t>(stream->Size() - sizeof(decompressedSize));
+		stream->Read(reinterpret_cast<char*>(buffer.data()), buffer.size());
+		auto decompressedBuffer = std::vector<uint8_t>(decompressedSize);
+		auto r =
+		    uncompress(decompressedBuffer.data(), reinterpret_cast<mz_ulong*>(&decompressedSize), buffer.data(), buffer.size());
```
bwrsandman, line 53: Fix warnings

bwrsandman (changes requested):

src/Game.cpp
bwrsandman: I would like to see all `*.zzz` files in `OutsideMeshes/` loaded here using the stem as the name.
See #378 with `*.raw` files

src/Resources/Loaders.cpp
raffclar: Fixed

bwrsandman (approved):
Fixed an include. Nice catch for "raw/" etc

bwrsandman:
Miniz adds warnings to the projects.
I would like it to be a static library instead of directly including sources.
vcpkg has miniz
I also wonder if minizip would work because it is a much more available library for linux system deps and vcpkg had it

raffclar:
I'll explore minizip and add a new commit to switch it over

bwrsandman (changes requested):
Will review when miniz.c is removed from source

.clang-format-ignore
```diff
@@ -3,3 +3,5 @@ apps/anmtool/json.hpp
 apps/l3dtool/tiny_gltf.h
 apps/l3dtool/json.hpp
 src/Common/stb_image_write.h
+src/3D/miniz.h
+src/3D/miniz.c
```
bwrsandman, line 7: You can remove these lines now. Keep a line break at the end of the file

.licenserc.json
```diff
@@ -12,6 +12,8 @@
     "ignore": [
         "externals/",
         "stb_image_write.h",
-        "tiny_gltf.h"
+        "tiny_gltf.h",
+        "miniz.h",
+        "miniz.c"
```
bwrsandman, line 17: ditto

src/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-file(GLOB_RECURSE OPENBLACK_SOURCES "${CMAKE_CURRENT_LIST_DIR}/*.cpp")
+file(GLOB_RECURSE OPENBLACK_SOURCES "${CMAKE_CURRENT_LIST_DIR}/*.cpp" "${CMAKE_CURRENT_LIST_DIR}/*.c")
```
bwrsandman, line 1: You can remove this change

bwrsandman (changes requested):

bwrsandman (approved):

PR #331, opened 2022-01-12
Refine the alpha threshold, add blending

raffclar:


assets/shaders/fs_object.sc
```diff
@@ -10,7 +10,8 @@ void main()
 	const vec3 lightColor = vec3(1.0f, 1.0f, 1.0f);
 	const vec4 lightPos = vec4(-4000.0f, 1300.0f, -1435.0f, 1.0f);
 	const float ambientStrength = 0.5f;
-	const float alphaThreshold = 1.0f / 255.0f;
+	// The alpha threshold different for each mesh
```
bwrsandman, line 13: Mark as a todo

bwrsandman:
I'll have to run and compare to approve. Don't have the correct set up right now but will update.

assets/shaders/fs_object.sc
raffclar: Done

bwrsandman (changes requested):
Very good change for the trees, could you elaborate on possible sources of threshold per mesh?

I would like to avoid the case of going back and forth between different threshold values because this isn't the first time we change this bit of the shader.

assets/shaders/fs_object.sc
```diff
@@ -10,7 +10,8 @@ void main()
 	const vec3 lightColor = vec3(1.0f, 1.0f, 1.0f);
 	const vec4 lightPos = vec4(-4000.0f, 1300.0f, -1435.0f, 1.0f);
 	const float ambientStrength = 0.5f;
-	const float alphaThreshold = 1.0f / 255.0f;
+	// TODO(raffclar): The alpha threshold is different for each mesh
+	const float alphaThreshold = 0.768627f;
```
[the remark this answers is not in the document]
raffclar: The current value was retrieved though Nsight by looking at the draw calls for a tree on Land 1. Different meshes can have different thresholds. There's no observable pattern for the threshold yet.

raffclar:
The tree that's being drawn with this alpha threshold:
image

bwrsandman:
With the discoveries of #334, I would say that these changes are insufficient.
Alpha threshold should only be used with specific render modes. The value of the threshold changes per primitive, so that value needs to be brought from the 5th byte of a primitive file to the draw call of the spans.

The shader should either be split or have logic that ignores alpha threshold when not used.

src/Renderer.cpp
```diff
@@ -436,8 +436,10 @@ void Renderer::DrawPass(const MeshPack& meshPack, const DrawSceneDesc& desc) con
 			// clang-format off
 			submitDesc.state = 0u
 				| BGFX_STATE_WRITE_MASK
-				| BGFX_STATE_DEPTH_TEST_LESS
-				| BGFX_STATE_MSAA
+				| BGFX_STATE_BLEND_FUNC(BGFX_STATE_BLEND_SRC_ALPHA, BGFX_STATE_BLEND_INV_SRC_ALPHA)
+				| BGFX_STATE_BLEND_EQUATION(BGFX_STATE_BLEND_EQUATION_ADD)
```
bwrsandman, line 440: These are dependent on render mode. See the table I made: https://github.com/openblack/openblack/wiki/Vanilla-Render-Modes

PR #333, opened 2022-01-12
Emulate vanilla FOV

raffclar:
Vanilla:
image
Us:
image

@bwrsandman thanks for this


bwrsandman (approved):
Looks good!

Don't forget to change the parameter names on the header as well

PR #335, opened 2022-01-15
Sky shader

bwrsandman:
Implement the sky texture interpolation as in fragment shader.
Make use of the time of day values for night, dusk and day.
Use that value for terrain brightness

https://user-images.githubusercontent.com/1013356/149610939-baa9642a-b44a-45a5-8433-475994856950.mp4


bwrsandman:
Added shading for objects and water according to time of day

image

raffclar (approved):

PR #336, opened 2022-01-17
Move project to C++20

bwrsandman:
- [x] Remove homebrew std::span implementation
- [x] Github actions for ubuntu must be updated to use g++ >= 10 (22.04 in 21st of April 2022)
    * stdc++-10 has span: https://ubuntu.pkgs.org/21.10/ubuntu-main-amd64/libstdc++-10-dev_10.3.0-11ubuntu1_amd64.deb.html
- [x] Update json.hpp and tinygltf with new c++20 compatible version
- [x] Deprecate old compilers that don't support c++20 if there are any
    - [x] stdc++-9 (Ubuntu 20.04 LTS)
- [x] Remove filesystem header check
- [x] Check if compatible with #262


bwrsandman:
Fairly certain the remaining fails are due to stale caches

bwrsandman:
It is indeed the cache. If I turn off the cache all tests pass: https://github.com/openblack/openblack/pull/336/commits/aef0a6e17ab5499b944bcc1e5855f86a72e13549

handsomematt (approved):
code changes look good to me, it's shit we were using already anyway already

PR #337, opened 2022-01-17
Physics scaling

bwrsandman:


bwrsandman:
We need a way to set static mobiles as not active in the simulation until they are interacted with.

https://user-images.githubusercontent.com/1013356/149793490-aeb768e1-fc24-4ecc-8b7e-6f01ad6b77b2.mp4

PR #330, opened 2022-01-11
Add names and types to info structs

bwrsandman:
This is done thanks to the info*.txt files on the mac disk

PR #285, opened 2021-03-08
Villager state machine framework + implementation of a few states

bwrsandman:
This PR will probably be split up into more logical changes once it has been prototyped and tested.

TODO:
- [x] #384
- [ ] Make more ECS?

PR #338, opened 2022-01-22
Pathfinding (Wall hug algorithm)

bwrsandman:
This PR adds pathfinding to all mobile wall hug. It was all done using ECS and is based on how vanilla works. The precision is higher than vanilla because of more pure calculations such as ray casting and circle-circle intersection and by using floating point numbers instead of fixed point.

https://user-images.githubusercontent.com/1013356/154280557-05b23b1d-f609-4ac9-b57e-de7653c69f78.mp4

https://user-images.githubusercontent.com/1013356/158028030-71709925-2054-4ebb-98e2-2f393dbbe99a.mp4

https://user-images.githubusercontent.com/1013356/158035188-cab3714e-53b0-4609-b00f-8f001b6d0b57.mp4

This PR is currently needs cleaning up:
- [x] Get some nice videos of villagers walking
    - [x] mobilewallhug1.txt
    - [x] mobilewallhug2.txt
- [x] Depends on #330
- [x] Depends on #265 
- [x] Depends on #275 
- [x] Depends on #358 for the unit tests
- [x] Depends on #381
- [x] Depends on #393 for tests loading scenes
- [x] Either depends or is a dependency of #285 (the go home state uses wall hug)\
- [x] #413 because the unit tests currently fail due to not getting the right mesh information from info.dat
- [x] Fix failing test cases
- [x] Integrate BBs into physics fixes for meshes without physics meshes (e.g. trees)
- [x] Remove hacks
- [x] Fix issue with hardcoded scale being needed in pathfinding algorithm
- [ ] Whole lot of testing

### Integration

Walking is done in discrete steps. Each turn, one step is performed and collision detection and state machine execution is done.

### State Machine

The way pathfinding works is with a state machine.
The states are:
* `ARRIVED`: Endpoint, no more transitions possible
* `FINAL_STEP`: Agent is within one step from goal and it just needs to go there in a straight line, no collision checking.
* `STEP_THROUGH`: Agent will walk through obstacles. Used to walk out of the inside of an obstacle.
* `LINEAR(Clockwise, Counter-clockwise or undefined)`: Agent will walk in a straight line until it either arrives at its goal or encounters an obstacle.
* `ORBIT(Clockwise or Counter-clockwise)`: Agent has reached an obstacle and will walk around it until it either encounters another obstacle or can walk in a straight line to goal again.
* `EXIT_CIRCLE(Clockwise or Counter-clockwise)`: In-between state from `ORBIT` to `LINEAR`

### Collision
Obstacles are calculated using 2d circles (or ellipses in some cases).

For linear movement, 2d raycasting is used to check for future collisions.
For orbital movement, the arc of a circle is checked by calculating the intersection of two circles (the one the agent is orbiting and every other in the region) and checking the closest intersection point for future collisions.


bwrsandman:
Can't include the tests because they rely on `.\DATA\LANDSCAPE\MULTI_PLAYER\MPM_2P_1.lnd`
Mock data is probably the solution to this.
Also CTests will not support additional command line arguments.

bwrsandman:
The current state is that the branch is getting mature and [f8753ee](https://github.com/openblack/openblack/pull/338/commits/f8753ee9ad4dc5998ff1fdd7432a855b586e6ff9) passes CI.
The last few commits don't but they're there for tests (which require game data or mock game data), leftover commented out code for corner cases and the hack for the demo.
We should be able to merge f8753ee9ad4dc5998ff1fdd7432a855b586e6ff9 and split the tests for later, or work on the old mock run tests and then add back the pathfinding tests.
I need to check if the tests actually still pass with game data.

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,634 @@ PathfindingSystem& PathfindingSystem::instance()
...
 	auto& registry = Game::instance()->GetEntityRegistry();
 
-	// ARRIVED:
+	// 1.  ARRIVED:
+	//         If AreWeThere is false, set to STEP_THROUGH (and it will trigger following steps)
 	registry.Each<const MoveStateArrivedTag, const Transform, const WallHug>(
-	    [](const MoveStateArrivedTag&, const Transform&, const WallHug&) {
-		    SPDLOG_LOGGER_WARN(spdlog::get("pathfinding"), "ARRIVED state not implemented in move to pathfinding");
+	    [&registry](entt::entity entity, const MoveStateArrivedTag& state, const Transform& transform, const WallHug& wallHug) {
+		    auto a = glm::xz(transform.position);
```
bwrsandman, line 459: This was debug, remove use of variable a

test/mobile_wall_hug/test_mobile_wall_hug.cpp
```diff
@@ -0,0 +1,319 @@
...
+			uint8_t turns_to_obstacle;
+			uint8_t field_0x5;  // TODO: Don't merge until this field is named
+			uint16_t field_0x6; // TODO: Don't merge until this field is named
+		};
+
+		uint32_t turn;
+		uint32_t id;
+		glm::vec2 pos;
+		uint16_t field_0x24; // TODO: Don't merge until this field is named
+		uint8_t field_0x26;  // TODO: Don't merge until this field is named
```
bwrsandman, line 107: These fields need to be resolved somehow

test/mobile_wall_hug/test_mobile_wall_hug.cpp
```diff
@@ -0,0 +1,322 @@
...
+		{
+			const auto& villager_comp = registry.Get<ecs::components::Villager>(villager_entt);
+			const auto& villager_transform = registry.Get<ecs::components::Transform>(villager_entt);
+			const auto& villager_wallhug = registry.Get<ecs::components::WallHug>(villager_entt);
+			bool villager_has_obstacle = registry.AnyOf<ecs::components::WallHugObjectReference>(villager_entt);
+			const auto& state = expected_states[turn - start_turn];
+			const auto msg = std::string("on turn ") + std::to_string(turn) + " in range " + std::to_string(start_turn) + "-" +
+			                 std::to_string(last_turn);
+
+			switch (state.move_state)
```
bwrsandman, line 224: Add default case

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,636 @@ PathfindingSystem& PathfindingSystem::instance()
...
+	const auto angle = glm::atan(diff.y, diff.x);
+	InitializeStep(transform, wallHug, angle);
+}
+
+void InitializeStepAroundObstacle(Transform& transform, WallHug& wallHug, const Fixed& obstacle, float numCirclesAway,
+                                  bool clockwise)
+{
+	// Use doubles for atan to have same precision on linux, osx and win32
+	const glm::dvec2 diff = obstacle.boundingCenter - glm::xz(transform.position);
+	const float angle = glm::atan(diff.y, diff.x);
```
bwrsandman, line 65: Explicit cast to float

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,636 @@ PathfindingSystem& PathfindingSystem::instance()
...
+	    {cellIndex.x - 1, cellIndex.y - 1}, // Over and left
+	    {cellIndex.x - 1, cellIndex.y + 1}, // Under and left
+	    {cellIndex.x + 1, cellIndex.y - 1}, // Over and right
+	    {cellIndex.x + 1, cellIndex.y + 1}, // Under and right
+	};
+}
+
+/// Iterate between all adjacent grids and find closest object that the ray (step) intersects with (circle)
+/// If that object is in front (and we are not in it) and less than 256 steps away, set as target and store steps
+bool LinearScanForObstacle(entt::entity entity, const glm::vec2& pos, const glm::vec2& dest, const glm::vec2& step)
```
bwrsandman, line 117: dest is unused

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,636 @@ PathfindingSystem& PathfindingSystem::instance()
...
+					const auto p2 = p0 + a * (p1 - p0) / d;       // midpoint of overlap
+					const auto diff = p1 - p0;
+					const auto difft = glm::vec2(diff.y, -diff.x); // 90 degree rotation
+					const auto p3 = p2 + h * difft / d;            // *** This is the correct one to take
+					const auto p4 = p2 - h * difft / d;
+
+					const auto v0 = p3 - obstacleFixed.boundingCenter; // ***
+					const auto v1 = p4 - obstacleFixed.boundingCenter;
+					const auto d0 = glm::length(v0);
+					const auto d1 = glm::length(v1);
```
bwrsandman, line 275: d1 and d2 are unused

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,636 @@ PathfindingSystem& PathfindingSystem::instance()
...
+						angle0 = 2.0f * glm::pi<float>() - angle0;
+					}
+					if ((cp1 > 0.0f) ^ clockwise)
+					{
+						angle1 = 2.0f * glm::pi<float>() - angle1;
+					}
+
+					const auto t0 = angle0 * 2.0f / 3.0f * obstacleFixed.boundingRadius * todoScaling / wallHug.speed; // ***
+					const auto t1 = angle1 * 2.0f / 3.0f * obstacleFixed.boundingRadius * todoScaling / wallHug.speed;
+					int t = glm::round(glm::min(t0, t1));
```
bwrsandman, line 296: Explicit cast to int

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,636 @@ PathfindingSystem& PathfindingSystem::instance()
...
+		    state.stepGoal = goal;
+	    },
+	    exclude...);
+}
+
+template <MoveState S>
+bool CellTransition(entt::entity entity, const MoveStateTagComponent<S> state, Transform& transform, WallHug& wallHug);
+
+template <>
+bool CellTransition(entt::entity entity, const MoveStateTagComponent<MoveState::Linear> state, Transform& transform,
```
bwrsandman, line 408: state is unused

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +38,636 @@ PathfindingSystem& PathfindingSystem::instance()
...
+					auto& fixed = registry.Get<const Fixed>(f);
+					const auto d2 = glm::distance2(fixed.boundingCenter, obstacleFixed.boundingCenter);
+					const auto r = fixed.boundingRadius + obstacleFixed.boundingRadius;
+					const auto r2 = r * r;
+					return d2 < r2 && d2 > 0.0f;
+				});
+				if (iter != e.cend())
+				{
+					// FIXME(bwrsandman): WTF why is scaling needed here and not in radius already????
+					const float todoScaling = 1.095f;
```
bwrsandman, line 252: What's up with this?

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +37,560 @@ PathfindingSystem& PathfindingSystem::instance()
...
+	const auto angle = glm::atan(diff.y, diff.x);
+	InitializeStep(transform, wallHug, angle);
+}
+
+void InitializeStepAroundObstacle(Transform& transform, WallHug& wallHug, const Fixed& obstacle, float numCirclesAway,
+                                  bool clockwise)
+{
+	// Use doubles for atan to have same precision on linux, osx and win32
+	const glm::dvec2 diff = obstacle.boundingCenter - glm::xz(transform.position);
+	const float angle = static_cast<float>(glm::atan(diff.y, diff.x));
```
bwrsandman, line 64: use auto

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -26,45 +37,560 @@ PathfindingSystem& PathfindingSystem::instance()
...
+	    });
+}
+
+// TODO: move_map_object is more complex than this. update to the map might be needed
+template <MoveState S, typename... Exclude>
+void ApplyStepGoal(ecs::Registry& registry, Exclude... exclude)
+{
+	// TODO: WallHug might be redundant here
+	registry.Each<const MoveStateTagComponent<S>, const WallHug, Transform>(
+	    [](const MoveStateTagComponent<S>& state, const WallHug& wallHug, Transform& transform) {
```
bwrsandman, line 365: wallhug is unused

apps/packtool/packtool.cpp
```diff
@@ -371,6 +402,52 @@ struct Arguments
...
+	{
+		range += ":1";
+		colCount = 1;
+	}
+	// If there is only one, it's followed by length and is like :currentSize:length
+	if (colCount == 1)
+	{
+		const auto pos = range.find(':');
+		range = range.substr(0, pos + 1) + std::to_string(currentSize) + range.substr(pos);
+		colCount = 1;
```
bwrsandman, line 425: Unused

bwrsandman:
Error is due to change in `.github/workflows/ci-vcpkg.yml` which is now required to be in upstream before it will be applied.
In previous runs it worked so this error can be safely ignored.

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -28,43 +39,557 @@ PathfindingSystem& PathfindingSystem::instance()
...
+	}
+
+	// Store object and number of steps away
+	registry.Assign<WallHugObjectReference>(entity, static_cast<decltype(WallHugObjectReference::stepsAway)>(numSteps),
+	                                        *fixedEntity);
+	return true;
+}
+
+/// If we have a number of turns to an obstacle but no obstacle saved: search in an arc to find one
+bool OrbitScanForObstacle(entt::entity entity, bool clockwise, Transform& transform, WallHug& wallHug)
```
raffclar, line 189: Clang-Tidy: Function 'OrbitScanForObstacle' is within a recursive call chain
[misc-no-recursion]

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -28,43 +39,557 @@ PathfindingSystem& PathfindingSystem::instance()
...
+/// +-----+-----+-----+
+/// | [2] | [0] | [1] |
+/// +-----+-----+-----+
+/// | [6] | [3] | [8] |
+/// +-----+-----+-----+
+std::array<ecs::Map::CellId, 9> GetNeighboringCells(const glm::vec2& pos)
+{
+	const auto& map = Game::instance()->GetEntityMap();
+	const auto cellIndex = map.GetGridCell(pos);
+	assert(glm::compMin(cellIndex) > 0 && glm::all(glm::lessThan(cellIndex, map._gridSize - glm::u16vec2(1))));
```
raffclar, line 100: Getting assertion failure with teleporting:

pos: {={={x=0, y=0}, ={r=0, g=0}, ={s=0, t=0}}}
cellIndex: {={={x=0, y=0}, ={r=0, g=0}, ={s=0, t=0}}}

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -28,43 +39,557 @@ PathfindingSystem& PathfindingSystem::instance()
...
+				}
+				else
+				{
+					// TODO:
+					// if intersect[0].obj is None:  # True
+					//     self.init_steps_xz()
+					//     # self.field_0x78 = 0x10
+					//     self.circle_hug_info.reset(self)
+					//     self.move_state = MoveState.STEP_THROUGH
+					// assert(false);
```
raffclar, line 303: Why is this in TODO?

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -28,43 +39,557 @@ PathfindingSystem& PathfindingSystem::instance()
...
+		    const auto positionId = Map::GetGridCell(position);
+		    const auto goalId = Map::GetGridCell(state.stepGoal);
+		    if (positionId != goalId)
+		    {
+			    CellTransition(entity, state, transform, wallHug);
+		    }
+	    });
+}
+
+// TODO: move_map_object is more complex than this. update to the map might be needed
```
raffclar, line 359: What is move_map_object? RE'd function?

raffclar (changes requested):

apps/packtool/packtool.cpp
```diff
@@ -326,10 +328,39 @@ int WriteRaw(const std::filesystem::path& outFilename, const std::vector<std::fi
 	return EXIT_SUCCESS;
 }
 
-int WriteMeshFile(const std::filesystem::path& outFilename) noexcept
+int WriteMeshFile(const std::filesystem::path& outFilename, const std::vector<std::filesystem::path> inFilenames) noexcept
```
handsomematt, line 331: not a const reference of the vector instead?

src/ECS/Systems/PathfindingSystem.cpp
```diff
@@ -28,43 +39,557 @@ PathfindingSystem& PathfindingSystem::instance()
 
 PathfindingSystem::PathfindingSystem() = default;
 
+namespace
+{
+
+void InitializeStep(Transform& transform, WallHug& wallHug, float angle)
+{
+	transform.rotation = glm::eulerAngleY(-angle - glm::radians(90.0f));
```
handsomematt, line 47: magic number? why's it offset to 90deg?

apps/packtool/packtool.cpp
bwrsandman: Good catch

src/ECS/Systems/PathfindingSystem.cpp
bwrsandman: Not just offset by 90 degrees but also reflected.
I have no answer to this. It's just how the math ended up working out.
Maybe it's the implementation of `glm::eulerAngleY` or the 2d pathfinding space is offset?

src/ECS/Systems/PathfindingSystem.cpp
bwrsandman: This is a carry-over from my python implementation that was based on RE

src/ECS/Systems/PathfindingSystem.cpp
bwrsandman: I'm not sure if this is needed. It's a corner case.
If we notice some villagers getting stuck. This might be where they get stuck...

src/ECS/Systems/PathfindingSystem.cpp
bwrsandman: Should we avoid recursive calls or disable the check?
This would be the only place there is one.
I could rework it as a limited loop.

raffclar (approved):
Tested. Very nice. Maybe a default movement speed of 1? I spent a minute scratching my head wondering why the villagers weren't moving 😄

PR #339, opened 2022-01-22
Update EnTT to 3.9.0

bwrsandman:
ecs: in 3.9.0, size<C>() is replaced by storage<C>().size()
https://github.com/skypjack/entt/releases/tag/v3.9.0

This fixes the System Deps CI test


handsomematt (approved):
Looks fine, only other port we use that got an update is cxxopts to 3.0.0.
I think vcpkg has started their versioning system now so we can version lock in manifest from now on if required.

PR #340, opened 2022-01-22
CI System Deps: Speed-up compilation

bwrsandman:
It dawned on me that the system deps are probably compiling single threaded.
On a new system, `MAKEFLAGS` is commented out.
This change to the regex makes sure to uncomment the line if it is commented out.

For reference ubuntu vcpkg config takes 2 minutes 25 seconds to compile

### DIFF
--- a/.github/workflows/ci-system-deps.yml
+++ b/.github/workflows/ci-system-deps.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install dependencies and configure system
         run: |
           pacman -Syu --noconfirm --noprogressbar --needed git cmake mesa libx11 fmt spdlog sdl2 glm bullet
-          sed -i "s|MAKEFLAGS=.*|MAKEFLAGS=-j$(nproc)|" /etc/makepkg.conf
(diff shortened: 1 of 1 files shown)


bwrsandman:
Obviously, this will have to be rebased on  #339, but if it works then the compilation of openblack should take less than the current 5 minutes.

.github/workflows/ci-system-deps.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Install dependencies and configure system
         run: |
           pacman -Syu --noconfirm --noprogressbar --needed git cmake mesa libx11 fmt spdlog sdl2 glm bullet
-          sed -i "s|MAKEFLAGS=.*|MAKEFLAGS=-j$(nproc)|" /etc/makepkg.conf
+          sed -i "s|#? ?MAKEFLAGS=.*|MAKEFLAGS=-j$(nproc)|" /etc/makepkg.conf
```
bwrsandman, line 18: should be `-E -i`

bwrsandman:
* Building bgfx.cmake (cache miss) has gone from 23m 21s to 12m 43s
* Building openblack has gone from 6m 54s to 3m 55s

PR #343, opened 2022-01-31
Forward declare optimizations

bwrsandman:
Cleanup and making use of forward declaration to lower memory usage and speed-up compilation.

### DIFF
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -8,11 +8,13 @@
  *****************************************************************************/
 
 #include "Camera.h"
+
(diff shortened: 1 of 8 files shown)


handsomematt (changes requested):
CI errors in release builds

bwrsandman:
Github actions seems to have bugged out on ` VCPKG CI / build (windows-latest, debug) (pull_request)`
Linker had an access violation and cancelling the build has not updated the status of the checks here.

bwrsandman:
I think it's a false negative that persists because of cache

PR #346, opened 2022-02-01
Update Cam to be more like Black & White

tino113:
Made camera more like Black & White
changes from FPV to Orbit camera
stripped Y component from forward movement
changed hand pulling method
included scroll to zoom
moved FPV controls to QERV
made more 'floaty'

### DIFF
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -456,7 +456,6 @@ void Camera::Update(std::chrono::microseconds dt)
 	glm::mat3 rotation = glm::transpose(GetViewMatrix());
 
 	// deal with hand pulling camera around
-	
 	float worldHandDist = 0.0f;
 	int sWidth, sHeight;
 	Game::instance()->GetWindow()->GetSize(sWidth, sHeight);
@@ -483,7 +482,6 @@ void Camera::Update(std::chrono::microseconds dt)
 			_handDragMult /= sHeight;
 		}
 		else if (!hit) { // still on screen but did not hit land
-			//worldHandDist = 3000.f;
 			_handDragMult += 0.001f; // speed up movement
 		}
 		else // if hand is off screen, culled or behind camera.
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -354,6 +354,15 @@ void Camera::handleKeyboardInput(const SDL_Event& e)
 			break;
 		}
 	}
+	else
+	{
+		_dv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_drv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_dwv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_dsv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_ddv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_duv = glm::vec3(0.0f, 0.0f, 0.0f);
+	}
 }
 
 void Camera::handleMouseInput(const SDL_Event& e)
@@ -415,6 +424,12 @@ void Camera::handleMouseInput(const SDL_Event& e)
 	{
 		if (!_mmouseIsDown)
 			_lmouseIsDown = true;
+		_dv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_drv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_dwv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_dsv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_ddv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_duv = glm::vec3(0.0f, 0.0f, 0.0f);
 	}
 	else if (e.type == SDL_MOUSEBUTTONUP && e.button.button == SDL_BUTTON(SDL_BUTTON_MIDDLE))
 	{
@@ -424,6 +439,12 @@ void Camera::handleMouseInput(const SDL_Event& e)
 	{
 		_mmouseIsDown = true;
 		_lmouseIsDown = false;
+		_dv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_drv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_dwv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_dsv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_ddv = glm::vec3(0.0f, 0.0f, 0.0f);
+		_duv = glm::vec3(0.0f, 0.0f, 0.0f);
 	}
 	if (e.type == SDL_MOUSEWHEEL)
 	{
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -405,11 +405,43 @@ void Camera::handleMouseInput(const SDL_Event& e)
 	}
 	if (e.type == SDL_MOUSEWHEEL)
 	{
-		auto movementSpeed = _movementSpeed * std::max(log(_position.y), 0.0f) + 1.0f;
+		auto movementSpeed = _movementSpeed * log(_position.y + 1);
 		if (e.wheel.y != 0) // scroll up or down
 		{
-			_velocity.z +=
-			    (((movementSpeed * e.wheel.y * abs(e.wheel.y) * 5 * _maxMovementSpeed) - _velocity.z) * _accelFactor);
+			float dist = 9999.0f;
+			auto& dynamicsSystem = Game::instance()->GetDynamicsSystem();
+			if (auto hit = dynamicsSystem.RayCastClosestHit(_position, GetForward(), 1e10f))
+			{
+				dist = glm::length(hit->first.position - _position);
+			}
+			if (e.wheel.y > 0.0f) // scrolling in
+			{ 
+				if (dist > 40.0f) // if the cam is far from the ground
+				{ 
+					_velocity.z +=
+						(((movementSpeed * e.wheel.y * abs(e.wheel.y) * 5 * _maxMovementSpeed) - _velocity.z) * _accelFactor);
+				}
+				else // if the cam is just over the ground
+				{ 
+					if (_rotation.x > glm::radians(-60.0f)) // rotation greater than -60 degrees
+					{
+						_rotVelocity.x += (((-e.wheel.y * 4.0f * _maxRotationSpeed) - _rotVelocity.x) * _accelFactor);
+					}
+					
+				}
+			}
+			else // scrolling out
+			{ 
+				if (dist <= 40.0f && _rotation.x < glm::radians(-50.0f))
+				{
+					_rotVelocity.x += (((-e.wheel.y * 4.0f * _maxRotationSpeed) - _rotVelocity.x) * _accelFactor);
+				}
+				else 
+				{
+					_velocity.z +=
+					(((movementSpeed * e.wheel.y * abs(e.wheel.y) * 5 * _maxMovementSpeed) - _velocity.z) * _accelFactor);
+				}
+			}
 		}
 		if (e.wheel.x != 0) // mouse wheel left or right
 		{
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -9,7 +9,6 @@
 
 #include "Camera.h"
 #include "3D/LandIsland.h"
-#include "ECS/Components/Transform.h"
 #include "ECS/Registry.h"
 #include "Game.h"
 
@@ -19,10 +18,6 @@
 #include <glm/gtx/norm.hpp>
 #include <glm/gtx/rotate_vector.hpp>
 #include <glm/gtx/vector_angle.hpp>
-
-// debug logging remove
-#include "glm/gtx/string_cast.hpp"
-#include <spdlog/spdlog.h>
 
 using namespace openblack;
 
@@ -335,7 +330,6 @@ void Camera::handleMouseInput(const SDL_Event& e)
 	else if (e.type == SDL_MOUSEBUTTONDOWN && e.button.button == SDL_BUTTON(SDL_BUTTON_LEFT))
 	{
 		_groundHeightStart = land.GetHeightAt(glm::vec2(_position.x, _position.z));
-		SPDLOG_LOGGER_INFO(spdlog::get("game"), "test");
 	}
 	else if (e.type == SDL_MOUSEBUTTONDOWN && e.button.button == SDL_BUTTON(SDL_BUTTON_MIDDLE))
 	{
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -45,6 +45,8 @@ Camera::Camera(glm::vec3 position, glm::vec3 rotation)
     , _mouseIsDown(false)
     , _mouseIsMoving(false)
     , _shiftHeld(false)
+	, _handScreenVec(0)
+	,_handDragMult(0.0f)
 {
 	FlyInit();
 }
@@ -454,8 +456,8 @@ void Camera::Update(std::chrono::microseconds dt)
 	glm::mat3 rotation = glm::transpose(GetViewMatrix());
 
 	// deal with hand pulling camera around
-	glm::ivec2 handScreenVec(0);
-	float handDist = 0.0f, worldHandDist = 0.0f;
+	
+	float worldHandDist = 0.0f;
 	int sWidth, sHeight;
 	Game::instance()->GetWindow()->GetSize(sWidth, sHeight);
 	if (_mouseIsDown)
@@ -475,26 +477,31 @@ void Camera::Update(std::chrono::microseconds dt)
 			SDL_GetMouseState(&mousePosition.x, &mousePosition.y);
 			glm::ivec2 handScreenCoords = glm::ivec2(handToScreen);
 			handScreenCoords.y = sHeight - handScreenCoords.y;
-			handScreenVec = mousePosition - handScreenCoords;
-			handDist = glm::length(glm::vec2(handScreenVec));
+			_handScreenVec = mousePosition - handScreenCoords;
+			_handDragMult = glm::length(glm::vec2(_handScreenVec));
 			worldHandDist = glm::length(hit->position - handPos);
-			handDist /= sHeight;
+			_handDragMult /= sHeight;
 		}
-		else // if hand is off screen, culled, behind camera or does not hit land.
+		else if (!hit) { // still on screen but did not hit land
+			//worldHandDist = 3000.f;
+			_handDragMult += 0.001f; // speed up movement
+		}
+		else // if hand is off screen, culled or behind camera.
 		{
 			_hVelocity = glm::vec3(0.f);
 		}
 	}
 	else
 	{
+		_handDragMult = (_handDragMult > 0.0f)? _handDragMult * 0.96f : _handDragMult; // if let go of mouse button slow down hand movement.
 		_hVelocity = glm::vec3(0.f);
 	}
 
-	if (handDist > 0.0f)
+	if (_handDragMult > 0.0f)
 	{
 		auto momentum = _position.y / 300;
-		auto forward = glm::normalize(GetForward() * glm::vec3(1.f, 0.f, 1.f)) * static_cast<float>(handScreenVec.y * momentum);
-		auto right = GetRight() * -static_cast<float>(handScreenVec.x * momentum);
+		auto forward = glm::normalize(GetForward() * glm::vec3(1.f, 0.f, 1.f)) * static_cast<float>(_handScreenVec.y * momentum);
+		auto right = GetRight() * -static_cast<float>(_handScreenVec.x * momentum);
 		auto futurePosition = _position + forward + right;
 		auto logPosY = log(_position.y + 1.0f);
 		auto handVelHeightMult = logPosY * logPosY;
@@ -507,8 +514,16 @@ void Camera::Update(std::chrono::microseconds dt)
 	if (worldHandDist > 2000.0f)
 	{
 		// speed up movement for really distant hand pulls.
-		handDist = (((handDist * 2 - 1) * (handDist * 2 - 1)) * -1) + 1;
+		_handDragMult += 0.001f;
 	}
+	else if (worldHandDist < 2000.0f && worldHandDist > 0.0f && _handDragMult >= 0.1f)
+	{
+		// slow movement for less distant hand pulls.
+		_handDragMult *= 0.98f;
+	} else if (worldHandDist < 0.0f)
+	{
+		_handDragMult = 0.0f;
+	}	
 
 	if (_flyInProgress)
 	{
@@ -563,7 +578,7 @@ void Camera::Update(std::chrono::microseconds dt)
 	else
 	{
 		_velocity += (((_dv * _maxMovementSpeed) - _velocity) * _accelFactor);
-		_hVelocity *= handDist;
+		_hVelocity *= _handDragMult;
 		_position += rotation * (_velocity + _hVelocity) * fdt;
 	}
 	const auto& land = Game::instance()->GetLandIsland();
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -45,8 +45,8 @@ Camera::Camera(glm::vec3 position, glm::vec3 rotation)
     , _mouseIsDown(false)
     , _mouseIsMoving(false)
     , _shiftHeld(false)
-	, _handScreenVec(0)
-	,_handDragMult(0.0f)
+    , _handScreenVec(0)
+    , _handDragMult(0.0f)
 {
 	FlyInit();
 }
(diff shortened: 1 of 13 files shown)


src/3D/Camera.cpp
```diff
@@ -172,39 +228,98 @@ void Camera::handleMouseInput(const SDL_Event& e)
...
 	{
-		glm::vec3 rot = GetRotation();
-
-		rot.y -= e.motion.xrel * _freeLookSensitivity * 0.1f;
-		rot.x -= e.motion.yrel * _freeLookSensitivity * 0.1f;
-
-		SetRotation(rot);
+		int width = 1920;
+		int height = 1080;
+		Game::instance()->GetWindow()->GetSize(width,height);
```
bwrsandman, line 233: Why reset the resolution? The resolution can be set from the command line if you don't like the current one.

src/3D/Camera.cpp
```diff
@@ -40,22 +45,32 @@ void Camera::SetProjectionMatrixPerspective(float xFov, float aspect, float near
...
 
+float Camera::GetGroundHeight() const
+{
+	float intersectDistance = 0.0f;
+	const auto rayDirection = glm::vec3(0.0f, -1.0f, 0.0f);
+	const auto plane_origin = glm::vec3(0.0f, 0.0f, 0.0f);
+	const auto plane_normal = glm::vec3(0.0f, 1.0f, 0.0f);
+	glm::intersectRayPlane(_position, rayDirection, plane_origin, plane_normal, intersectDistance);
+	return intersectDistance;
+}
```
bwrsandman, line 72: Ground height is in function of the land heighmap. I thought we recently implemented this using bullet and intersection with the actual geometric.
This virtual plane is definitely wrong for getting the height.
If you mean to get the altitude, the function name should reflect it better.

bwrsandman (changes requested):

bwrsandman:
@raffclar did some recent changes to camera so he can add some input.

I am looking forward to zoom with scroll wheel and orbit camera though!

bwrsandman:
As for the formatting errors, we have clang-format support and if you build that target, it will format everything nicely as long as you have it installed.

The Github action might have a patch file for you too.

src/3D/Camera.cpp
tino113: It's not resetting the resolution, I'm just initializing two variables to be immediately updated by the GetSize function, we need the width and height to calculate how much the yaw and pitch are dependent on the width and height.

src/3D/Camera.cpp
tino113: this function is not named particularly well, yes altitude of the camera from the ground plane. The function you refer to appears to take an X and Y position and returns the altitude of the land at that coordinate, this is useful but I want to know here how high above the land the camera is, raytracing with the actual geom is overkill and will take a lot of time, for this calculation a ground plane is sufficient. If there's a way to integrate the two functions that could be useful. I'll have another look at this.

tino113:
oops, just realized I only committed one file, there should be changes to Camera.h and Game.cpp too, I'll make some adjustments and then commit again.

src/3D/imgui.ini
```diff
@@ -0,0 +1,5 @@
+[Window][Debug##Default]
+Pos=60,60
+Size=400,400
+Collapsed=0
+
```
bwrsandman, line 5: Why this file?

src/3D/Camera.cpp
```diff
@@ -131,7 +146,7 @@ void Camera::ProcessSDLEvent(const SDL_Event& e)
 {
 	if (e.type == SDL_KEYDOWN || e.type == SDL_KEYUP)
 		handleKeyboardInput(e);
-	else if (e.type == SDL_MOUSEMOTION || e.type == SDL_MOUSEBUTTONDOWN || e.type == SDL_MOUSEBUTTONUP)
+	else if (e.type == SDL_MOUSEMOTION || e.type == SDL_MOUSEBUTTONDOWN || e.type == SDL_MOUSEBUTTONUP || SDL_MOUSEWHEEL)
```
bwrsandman, line 149: `e.type == SDL_MOUSEWHEEL`

src/3D/Camera.cpp
bwrsandman: Ah yes. You don't need to give these values. It can become confusing if you do that.

src/3D/Camera.h
```diff
@@ -26,11 +26,22 @@ class Camera
...
+	    , _groundHeight(0.0f)
+	    , _groundHeightStart(0.0f)
+	    , _movementSpeed(4.0f)
+	    , _accelFactor(0.001f)
+	    , _rotAccelFactor(0.001f)
 	    , _maxMovementSpeed(0.005f)
+	    , _maxRotationSpeed(0.005f)
 	    , _freeLookSensitivity(1.0f)
 	{
 	}
```
bwrsandman, line 47: I know you said you're a little rusty in C++, but I think this is a good opportunity to split declaration and definition of the camera constructor.

What I mean is that we have this in the header:
```c++
Camera(glm::vec3 position, glm::vec3 rotation);
```
and this in the .cpp implementation:
```c++
Camera::Camera(glm::vec3 position, glm::vec3 rotation)
	    : _position(position)
	    , _viewCenterLand(_position)
	    , _rotation(glm::radians(rotation))
	    , _dwv(0.0f, 0.0f, 0.0f)
	    , _dv(0.0f, 0.0f, 0.0f)
	    , _drv(0.0f, 0.0f, 0.0f)
	    , _projectionMatrix(1.0f)
	    , _wVelocity(0.0f, 0.0f, 0.0f)
	    , _velocity(0.0f, 0.0f, 0.0f)
	    , _rotVelocity(0.0f, 0.0f, 0.0f)
	    , _groundHeight(0.0f)
	    , _groundHeightStart(0.0f)
	    , _movementSpeed(4.0f)
	    , _accelFactor(0.001f)
	    , _rotAccelFactor(0.001f)
	    , _maxMovementSpeed(0.005f)
	    , _maxRotationSpeed(0.005f)
	    , _freeLookSensitivity(1.0f)
	{
	}
```

src/3D/Camera.cpp
bwrsandman: Might even be the time to move to a switch statement:

```c++
switch (e.type)
{
    case SDL_KEYDOWN:
    case SDL_KEYUP:
        handleKeyboardInput(e);
        break;
    case SDL_MOUSEMOTION :
    case SDL_MOUSEBUTTONDOWN :
    case SDL_MOUSEWHEEL:
        handleMouseInput(e);
        break;
        break;
}
```

src/3D/Camera.cpp
```diff
@@ -172,39 +228,106 @@ void Camera::handleMouseInput(const SDL_Event& e)
...
+	{
+		_groundHeightStart = GetGroundPlaneAltitude();
+	}
+	else if (e.type == SDL_MOUSEBUTTONDOWN && e.button.button == SDL_BUTTON(SDL_BUTTON_MIDDLE))
+	{
+		// get the _viewCenterLand by raycasting to the land down the camera ray
+		float intersectDistance = 0.0f;
+		glm::vec3 viewVec = GetForward();
+		if (glm::intersectRayPlane(_position, viewVec, glm::vec3(0.0f, 0.0f, 0.0f), glm::vec3(0.0f, 1.0f, 0.0f),
+		                           intersectDistance))
```
bwrsandman, line 282: This is very similar to `GetGroundPlaneAltitude`, maybe you can make it multipurpose.

src/3D/Camera.cpp
bwrsandman: It's not super useful to call `glm::intersectRayPlane` with constant parameters.
For example, with plane_origin and plane_normal, we can simplify [`glm::intersectRayPlane`](https://github.com/g-truc/glm/blob/master/glm/gtx/intersect.inl#L5-L26) to be.
```c++
		if (glm::abs(rayDirection.y) > std::numeric_limits<float>::epsilon())
		{
			float intersectDistance = -_position.y / rayDirection .y;
			if (intersectDistance > 0.0f) {
				return intersectDistance;
			}
		}
		return 0.0f; // invalid value
```

In fact, that was in case you choose to make `rayDirection` variable. If it isn't, this is the function simplified

```c++
			if (_position.y > 0.0f) {
				return _position.y;
			}
            else
            {
                return 0.0f;  // invalid value
            }
```

bwrsandman:
You can ignore the `build (windows-latest, debug)` fail, it seems to be a CI false negative.

src/3D/Camera.h
bwrsandman: Oh and make sure the order of initialization is the same as declaration (see the warnings in diff)

src/Game.cpp
```diff
@@ -161,6 +164,7 @@ bool Game::ProcessEvents(const SDL_Event& event)
 		switch (event.key.keysym.sym)
 		{
 		case SDLK_ESCAPE:
+			// TODO: esc un-maximises if fullscreen
```
bwrsandman, line 167: No, escape quits the game, later it will bring up the menu.

src/3D/Camera.cpp
tino113: good point, this is just the y component of the camera position! over-complicating things..

src/3D/imgui.ini
tino113: hmm I've no idea, something in the build generated it, I'm not sure what it is..

src/3D/Camera.cpp
tino113: yes good point I'll change this.

src/3D/Camera.cpp
tino113: removed all references to this function in favor of:
`land.GetHeightAt(glm::vec2(_position.x, _position.z))`

tino113: which is the actual land height from your previous function

src/3D/Camera.cpp
tino113: did not change to vec2, just removed unnecessary initialization

src/3D/Camera.h
tino113: I think I've done this right, you will need to check.

src/3D/Camera.cpp
tino113: As I've now got rid of `GetGroundPlaneAltitude` I've left this in as a stand alone call to `glm::intersectRayPlane`

src/3D/Camera.cpp
tino113: nope, actually none of that is necessary, just using `_position.y` is more than sufficient. some issues around this, fixing now..

src/3D/Camera.cpp
```diff
@@ -189,47 +209,32 @@ void Camera::handleKeyboardInput(const SDL_Event& e)
...
-	}
-	else if (e.key.keysym.scancode == SDL_SCANCODE_R)
-	{
+	case SDL_SCANCODE_R:
 		_drv.x += (e.type == SDL_KEYDOWN) ? _movementSpeed : -_drv.x;
-	}
-	else if (e.key.keysym.scancode == SDL_SCANCODE_V)
-	{
+	case SDL_SCANCODE_V:
 		_drv.x += (e.type == SDL_KEYDOWN) ? -_movementSpeed : -_drv.x;
```
bwrsandman, line 242: Be careful, without a break statement, the execution will flow through and execute the next line in the switch statement.
https://docs.microsoft.com/en-us/cpp/cpp/break-statement-cpp?view=msvc-170

src/3D/Camera.cpp
```diff
@@ -21,6 +21,25 @@
...
+	_projectionMatrix = glm::mat4(1.0f);
+	_velocity = glm::vec3(0.0f, 0.0f, 0.0f);
+	_rotVelocity = glm::vec3(0.0f, 0.0f, 0.0f);
+	_groundHeight = 0.0f;
+	_groundHeightStart = 0.0f;
+	_accelFactor = 0.001f;
+	_movementSpeed = 4.0f;
+	_maxMovementSpeed = 0.005f;
+	_maxRotationSpeed = 0.005f;
+}
```
bwrsandman, line 41: The initializer list was better. Just that the order had to match the declaration

tino113:
@bwrsandman I think it's done, check you're happy.. the camera movement feels nice to me now and reminiscent of B&W. Still an issue with the hand grabbing "overshooting" but this is a more complex issue to solve for a later date.

bwrsandman:
Thanks for you patience. It looks a lot better now. I will want to test this on a build later.

src/3D/Camera.cpp
```diff
@@ -141,70 +194,186 @@ void Camera::handleKeyboardInput(const SDL_Event& e)
...
+			_dv += temp;
+			_duv = temp;
+		}
+		else
+		{
+			_dv -= _duv;
+			_duv = glm::vec3(0.0f, 0.0f, 0.0f);
+		}
+		/*_dv += (e.type == SDL_KEYDOWN) ? movementSpeed * glm::vec3(0.0f, 1.0f, 0.0f)
+		                               : -movementSpeed * glm::vec3(0.0f, 1.0f, 0.0f);*/
```
bwrsandman, line 269: Looks like you forgot to remove this

src/imgui.ini
```diff
@@ -0,0 +1,5 @@
+[Window][Debug##Default]
+Pos=60,60
+Size=400,400
+Collapsed=0
+
```
bwrsandman, line 5: You keep accidentally adding this. Might I suggest adding this line to the gitignore

bwrsandman: Just add this to .gitignore
```
imgui.ini
```

src/3D/Camera.cpp
tino113: done

tino113:
> Thanks for you patience. It looks a lot better now. I will want to test this on a build later.

Thanks for your patience! look forward to it.

bwrsandman:
Still have imgui.ini from your previous commit

bwrsandman:
First thing. This is a very nice effect. Definitely feels more floaty.

Are any of these what you mean by overshooting?

Stopping mouse after moving:

https://user-images.githubusercontent.com/1013356/152253497-5b62b31a-d4b3-4576-882a-6bf5cc36ea1e.mp4

Successive small movements:

https://user-images.githubusercontent.com/1013356/152253824-5246b0d4-8ce2-43cb-beb2-63a25816cdb4.mp4

tino113:
yes, this is what I mean. I tried to solve it by getting the world position of both the hand and the mouse, checking the distance and multiplying the movement with the distance. So that when the distance is small the change in velocity would be reduced to zero, however this only happened on actual mouse events, so for it to work you have to keep moving the mouse. So it cant be in an event, needs to be somewhere in the game loop as a separate part of the logic.

The way I see it, and the feeling I get from B&W is that when the mouse on screen is approaching the hand's position, camera movement slows and eventually approaches a stop, that's what we want to emulate. It's not a trivial issue to solve I think. If you're happy with everything else for now I'd say merge the rest and put this as a bug to solve. Will have to give some thought as to how to do this in an efficient manner.

tino113:
I've just thought, there's camera update, I can do it it in there and move the logic out of mouse move.. give me a bit I think I have a solution.

tino113:
I think it's done...

src/3D/Camera.cpp
```diff
@@ -110,101 +154,277 @@ bool Camera::ProjectWorldToScreen(const glm::vec3 worldPosition, const glm::vec4
...
-	// ignore all repeated keys
-	if (e.key.repeat > 0)
+	if (e.key.repeat > 0) // ignore all repeated keys
 		return;
 
-	if (e.key.keysym.scancode == SDL_SCANCODE_W)
+	auto movementSpeed = _movementSpeed * std::max(_position.y * 0.01f, 0.0f) + 1.0f;
+
+	switch (e.key.keysym.scancode)
 	{
```
bwrsandman, line 195: You'll need to fix these warnings by probably adding this at the end
```
default:
    break;
```

src/3D/Camera.cpp
```diff
@@ -110,101 +154,277 @@ bool Camera::ProjectWorldToScreen(const glm::vec3 worldPosition, const glm::vec4
...
+	case SDL_SCANCODE_V:
+		_drv.x += (e.type == SDL_KEYDOWN) ? -_movementSpeed : -_drv.x;
+		break;
 	}
-	else if (e.key.keysym.scancode == SDL_SCANCODE_S)
+}
+void Camera::handleMouseInput(const SDL_Event& e)
+{
+	const auto& land = Game::instance()->GetLandIsland();
+
```
bwrsandman, line 290: land is unused and is causing a warning that we didn't have before

bwrsandman (approved):
This is already much much better.

There are still corner cases to improve but they are not going to block for me.
Please have a look at the remaining warnings.

Infinite movement when hand is above the horizon:

https://user-images.githubusercontent.com/1013356/152367911-4d3217ff-1559-4410-8d6b-1fb14ea211c2.mp4

Attenuation is a bit slow when hand goes far away:


https://user-images.githubusercontent.com/1013356/152368126-d0138af2-84be-4b31-a92a-39e9e3c2a378.mp4

It's possible that this is vanilla behaviour

tino113:
> Infinite movement when hand is above the horizon:
looking back at B&W there appears to be a maximum movement distance
> Attenuation is a bit slow when hand goes far away
attenuation is done by distance in screen space coordinates, distance on the world should make no distance, but I'll have a look.

src/3D/Camera.cpp
```diff
@@ -9,28 +9,97 @@
...
 #include "3D/LandIsland.h"
+#include "ECS/Components/Transform.h"
 #include "ECS/Registry.h"
 #include "Game.h"
 
 #include <glm/gtx/euler_angles.hpp>
+#include <glm/gtx/intersect.hpp>
+#include <glm/gtx/norm.hpp>
+
+#define OUT
```
bwrsandman, line 20: ?

tino113: stylistic choice to make out params very visible in the function declaration, can be removed if you really dislike

src/3D/Camera.cpp
```diff
@@ -9,28 +9,97 @@
...
-	return getRotationMatrix() * glm::translate(glm::mat4(1.0f), -_position);
+	return GetRotationMatrix() * glm::translate(glm::mat4(1.0f), -_position);
 }
 
 glm::mat4 Camera::GetViewProjectionMatrix() const
 {
 	return GetProjectionMatrix() * GetViewMatrix();
 }
 
+bool Camera::RaycastCamToLand(OUT glm::vec3* hit)
```
tino113, line 62: like this, just looks good imo

src/3D/Camera.cpp
```diff
@@ -9,28 +9,97 @@
...
+		*hit = _position + viewVec * 100.f;
+	}
+	return false;
+}
+
+bool Camera::RaycastMouseToLand(OUT glm::vec3* hit)
+{
+	// get the hit by raycasting to the land down via the mouse
+	float intersectDistance = 0.0f;
+	int sWidth, sHeight;
```
raffclar, line 84: These are not initialised but used below with the vec4. Causing an exception on my end.

raffclar (changes requested):

src/3D/Camera.cpp
bwrsandman: We don't use this style anywhere else and seems kind of arbitrary to start doing it now just here.
Not to mention it's not a good place to add global defines.
`OUT` is also a very 1980s indicator in c++. Might be some modern compiler hints that do this.

bwrsandman: On another note, it might be better to have that parameter be a reference. This would add a linguistic guarantee that there is valid data there.
With a naked pointer, there are more chances of runtime errors.

bwrsandman: If you're looking for a nice api, might I suggest returning an `std::optional<glm::vec3>`

src/3D/Camera.cpp
bwrsandman: Warnings are also picking it up as potential bugs

tino113: this was a good catch, they are not initialized at all! I had forgotten a line!

src/3D/Camera.cpp
tino113: good point I've swapped this to be pass by reference rather than the pointer.

regards `std::optional<glm::vec3` that's for the return type? what about if I want to use it to qualify if the ray actually hits? how would this work?

bwrsandman: Optional as a return type. Instead of returning false you would return an `std::nullopt` else you would return the vec3
https://en.cppreference.com/w/cpp/utility/optional

src/3D/Camera.cpp
```diff
@@ -110,101 +178,287 @@ bool Camera::ProjectWorldToScreen(const glm::vec3 worldPosition, const glm::vec4
...
+	{
+		// get hand transform and project to screen coords
+		auto& entityReg = Game::instance()->GetEntityRegistry();
+		auto handEntity = Game::instance()->GetHand();
+		auto& handTransform = entityReg.Get<ecs::components::Transform>(handEntity);
+		auto handPos = handTransform.position;
+		glm::vec3 handToScreen;
+		glm::vec4 viewport = glm::vec4(0, 0, sWidth, sHeight);
+		glm::vec3 hit;
+		if (ProjectWorldToScreen(handPos, viewport, handToScreen) and RaycastMouseToLand(hit))
```
bwrsandman, line 399: This isn't Python ;)

Use `&&`

tino113:
logic of mouse raycast is wrong, will need to update, just discovered while doing the camera flight code.

src/3D/Camera.cpp
tino113: there's a python wrapper for BGFX maybe we should switch the entire thing to python so I can use `and`

src/3D/Camera.cpp
tino113: Ok I've changed it somewhat to this and also returning an optional transform rather than just a position (I also want the normals for the camera path)

tino113:
#347 tackled somewhat in the last commit, flight curve is not the same as B&W will likely need adjusting when bookmarks are implemented.

src/3D/Camera.cpp
```diff
@@ -9,28 +9,136 @@
...
-	return getRotationMatrix() * glm::translate(glm::mat4(1.0f), -_position);
+	return GetRotationMatrix() * glm::translate(glm::mat4(1.0f), -_position);
 }
 
 glm::mat4 Camera::GetViewProjectionMatrix() const
 {
 	return GetProjectionMatrix() * GetViewMatrix();
 }
 
+const std::optional<ecs::components::Transform> Camera::RaycastCamToLand()
```
bwrsandman, line 64: A note about returning const (not not const pointers). They don't prevent the caller from storing the result in a non-const copy. Therefore returning const is a useless qualifier.

On another note, it is good practice to set getters and functions that do not modify `this` as const. It gives the compiler hints for better optimization. 
https://www.geeksforgeeks.org/const-member-functions-c/

src/3D/Camera.cpp
```diff
@@ -9,28 +9,136 @@
...
+		intersectionTransform.position = _position + viewVec * intersectDistance;
+		intersectionTransform.rotation = glm::mat3(1.0f);
+		return std::make_optional(intersectionTransform);
+	}
+	// else // point 100 away from camera to rotate around etc. not currently used
+	//{
+	// intersectionTransform.position = _position + viewVec * 100.f;
+	// intersectionTransform.rotation = glm::mat3(1.0f);
+	// return std::make_optional(intersectionTransform);
+	//}
```
bwrsandman, line 88: Are you sure you didn't mean to delete this bit of code?
Is it a To-do?

src/3D/Camera.cpp
```diff
@@ -9,28 +9,136 @@
...
+}
+
+const std::optional<ecs::components::Transform> Camera::RaycastMouseToLand()
+{
+	// get the hit by raycasting to the land down via the mouse
+	ecs::components::Transform intersectionTransform;
+	float intersectDistance = 0.0f;
+	int sWidth, sHeight;
+	Game::instance()->GetWindow()->GetSize(sWidth, sHeight);
+	glm::vec4 viewport = glm::vec4(0, 0, sWidth, sHeight);
```
bwrsandman: Remove this line

src/3D/Camera.cpp
```diff
@@ -9,28 +9,136 @@
...
+	_flyThreshold = 50.0f;
+	_flyFromPos = glm::vec3(0.0f, 0.0f, 0.0f);
+	_flyFromTan = glm::vec3(0.0f, 0.0f, 0.0f);
+	_flyToPos = glm::vec3(0.0f, 0.0f, 0.0f);
+	_flyToNorm = glm::vec3(0.0f, 0.0f, 0.0f);
+	_flyToTan = glm::vec3(0.0f, 0.0f, 0.0f);
+	_flyPrevPos = glm::vec3(0.0f, 0.0f, 0.0f);
+}
+
+glm::vec3 Camera::Fly(glm::vec3& fromPos, glm::vec3& fromTan, glm::vec3& toPos, glm::vec3& toTan, float t)
```
bwrsandman, line 136: You should const qualify the ref arguments that aren't meant to change.

Can you also explain what this function does? It's not very clear from the name and body.

src/3D/Camera.h
```diff
@@ -45,6 +37,11 @@ class Camera
 	[[nodiscard]] const glm::mat4& GetProjectionMatrix() const { return _projectionMatrix; }
 	[[nodiscard]] virtual glm::mat4 GetViewProjectionMatrix() const;
 
+	[[nodiscard]] const std::optional<ecs::components::Transform> RaycastCamToLand();
+	[[nodiscard]] const std::optional<ecs::components::Transform> RaycastMouseToLand();
+	[[nodiscard]] void FlyInit();
```
bwrsandman: Remove `[[nodiscard]] ` for void functions

bwrsandman (changes requested):

src/3D/Camera.h
```diff
@@ -45,6 +37,11 @@ class Camera
 	[[nodiscard]] const glm::mat4& GetProjectionMatrix() const { return _projectionMatrix; }
 	[[nodiscard]] virtual glm::mat4 GetViewProjectionMatrix() const;
 
+	[[nodiscard]] const std::optional<ecs::components::Transform> RaycastCamToLand();
+	[[nodiscard]] const std::optional<ecs::components::Transform> RaycastMouseToLand();
```
bwrsandman, line 41: Move const to end of the line.
Also these functions aren't called, can you provide an example of how they would be used?
Trying to avoid adding dead code

src/3D/Camera.cpp
tino113: deleted

src/3D/Camera.h
tino113: `RaycastCamToLand()` is a hangover from when I had the camera rotating around the center of the camera view for orbit rotation, it's now been deleted.

`RaycastMouseToLand()` is used to determine where a camera flight path should end and is used twice further down.

src/3D/Camera.cpp
tino113: deleted and code moved out as used only once.

src/3D/Camera.cpp
tino113: realized it cant be const anyway as it modifies the state of `this`

tino113:
My Ninja fails on linking so I can't test that build on my end, I've been using VS19 instead for now, hence why I'm not catching all the errors, I'll have to fix the linking issue. I hope this commit is ok. Sorry for my poor code consistency.

bwrsandman (approved):
LGTM

bwrsandman (changes requested):
There is definitely a regression. It was better prior to the latest changes.

Currently if you click, you move forwards a little bit.
Second, it's now very hard to move because as soon as the mouse goes above the horizon, it stops moving instead of decelerating.

tino113:
> There is definitely a regression. It was better prior to the latest changes.
> 
> Currently if you click, you move forwards a little bit. Second, it's now very hard to move because as soon as the mouse goes above the horizon, it stops moving instead of decelerating.

fixed the horizon issue, now will gradually accelerate, presuming that if you hold over the horizon you want to go further, if you let go it now gradually reduces speed.

regards camera flight movement, double clicking from far away moves you most of the way, then most of the way again, when you're approximately 20 world units away it stops taking you any further forward. Which is similar to the way B&W works, or is this not what you mean?

tino113:
2 steps forward one step back

bwrsandman:
Issues persisting are:
* Simply grabbing (left click) moves you forward
* Infinite scrolling beyond the horizon
* Formatting

bwrsandman (approved):
This is really good

https://user-images.githubusercontent.com/1013356/154285149-659399d9-1ecb-493c-9acc-572de5ab63fc.mp4

.

raffclar (approved):

PR #366, opened 2022-04-14
ci: Upload artifacts of builds

bwrsandman:
Each build will now provide artifacts in the summary of `VCPKG CI`, `Cross Compile CI` and `System Deps CI`.
We can provide these artifacts to users who don't know how to build, we can use them for tagged releases and we can also use them to quickly test a PR.

There may be a way to combine windows Debug and Release artifacts by giving them the same artifact name which will give an artifact with `Debug/` and `Release/` but that would require too much logic in the yaml file and I didn't do this way.

Screenshot from 2022-04-14 09-09-48


handsomematt (approved):
love how simple this is, can't wait to do something similar for a tagged release workflow

bwrsandman:
I decided to try and merge Debug and Release for the artifacts.

PR #367, opened 2022-04-14
Restore windows-2019 tests

bwrsandman:
According to https://github.com/actions/virtual-environments, windows-2022 and windows-latest are now the same


handsomematt (approved):
👍 lgtm depending on ci

PR #371, opened 2022-04-17
Fix std::vector out-of-bounds access when constructing std::span objects

raffclar:


raffclar:
Working locally with msvc. No out-of-bounds / access violations with this.

raffclar:
@bwrsandman Can you check when you have a sec

bwrsandman (approved):
Looks fine. I assume you've tested

raffclar:
> Looks fine. I assume you've tested

Windows only. Don't see how it could fail now though

Issue #211, opened 2020-02-08
Windowed mode resizing window


SulpherStaer:
![image](https://user-images.githubusercontent.com/5037244/74087722-50869f00-4a8f-11ea-8316-e1c9fc028cce.png)
Resizing the window does not resize the viewport, and the upper menu also vanishes.

FalseIlyu:
This one was solved with #373 ?

FalseIlyu:
I believe this issue has been solved with #373 so i'm closing it;

PR #221, opened 2020-02-13
Formatting and clean-up with clang tidy tests

bwrsandman:
I discovered that we have a `.clang-tidy`. I made sure we enforce it in our tests and added `clang-analyzer-*` too.

PR #373, opened 2022-04-19
Reset bgfx and camera projection on window resize

### LINKED ISSUE #100: Reset projection matrix on window resize

captnfab:
On window resize events, perform a [bgfx::reset](https://bkaradzic.github.io/bgfx/bgfx.html) to reset framebuffers, and reset camera projection.

Fixes #100, #211

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -179,6 +179,16 @@ bool Game::ProcessEvents(const SDL_Event& event)
 		{
 			return false;
 		}
+		else if (event.window.event == SDL_WINDOWEVENT_SIZE_CHANGED)
+		{
+			uint16_t width = static_cast<uint16_t>(event.window.data1);
+			uint16_t height = static_cast<uint16_t>(event.window.data2);
+			_renderer->Reset(width, height);
+			_renderer->ConfigureView(graphics::RenderPass::Main, width, height);
+
+			auto aspect = _window->GetAspectRatio();
+			_camera->SetProjectionMatrixPerspective(_config.cameraXFov, aspect, _config.cameraNearClip, _config.cameraFarClip);
+		}
 		break;
 	case SDL_KEYDOWN:
 		switch (event.key.keysym.sym)
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -392,7 +392,7 @@ bool Game::Run()
 	// create our camera
 	_camera = std::make_unique<Camera>();
 	auto aspect = _window ? _window->GetAspectRatio() : 1.0f;
-	_camera->SetProjectionMatrixPerspective(70.0f, aspect, 1.0f, 65536.0f);
+	_camera->SetProjectionMatrixPerspective(_config.cameraXFov, aspect, _config.cameraNearClip, _config.cameraFarClip);
 
 	_camera->SetPosition(glm::vec3(1441.56f, 24.764f, 2081.76f));
 	_camera->SetRotation(glm::vec3(0.0f, -45.0f, 0.0f));
--- a/src/Gui/Gui.cpp
+++ b/src/Gui/Gui.cpp
@@ -209,6 +209,12 @@ bool Gui::ProcessEventSdl2(const SDL_Event& event)
 		io.KeySuper = ((SDL_GetModState() & KMOD_GUI) != 0);
 		return io.WantCaptureKeyboard;
 	}
+	case SDL_WINDOWEVENT:
+		if (event.window.event == SDL_WINDOWEVENT_SIZE_CHANGED)
+		{
+			return false;
+		}
+		break;
 	}
 	return io.WantCaptureMouse;
 }
--- a/src/Game.h
+++ b/src/Game.h
@@ -137,6 +137,10 @@ public:
 		float bumpMapStrength {1.0f};
 		float smallBumpMapStrength {1.0f};
 
+		float cameraXFov {70.0f};
+		float cameraNearClip {1.0f};
+		float cameraFarClip {static_cast<float>(0x10000)};
+
 		bool bgfxDebug {false};
 		bool bgfxProfile {false};
 		bool running {false};
--- a/src/Renderer.cpp
+++ b/src/Renderer.cpp
@@ -135,11 +135,12 @@ Renderer::Renderer(const GameWindow* window, bgfx::RendererType::Enum rendererTy
 		window->GetNativeHandles(init.platformData.nwh, init.platformData.ndt);
 	}
 
-	init.resolution.reset = BGFX_RESET_NONE;
+	_bgfxReset = BGFX_RESET_NONE;
 	if (vsync)
 	{
-		init.resolution.reset |= BGFX_RESET_VSYNC;
+		_bgfxReset |= BGFX_RESET_VSYNC;
 	}
+	init.resolution.reset = _bgfxReset;
 	init.callback = dynamic_cast<bgfx::CallbackI*>(_bgfxCallback.get());
 
 #ifdef __APPLE__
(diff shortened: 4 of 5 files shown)


src/Game.cpp
```diff
@@ -179,6 +179,15 @@ bool Game::ProcessEvents(const SDL_Event& event)
 		{
 			return false;
 		}
+		else if (event.window.event == SDL_WINDOWEVENT_SIZE_CHANGED)
+		{
+			int width, height;
```
bwrsandman, line 184: I know we do it in other places but for readability, don't declare two variables on the same line.
If you really want one line try this:
```
glm::ivec2 resolution;
_window->GetSize(resolution,x, resolution.y);
```

src/Game.cpp
```diff
@@ -179,6 +179,15 @@ bool Game::ProcessEvents(const SDL_Event& event)
...
+			_window->GetSize(width, height);
+			auto aspect = _window->GetAspectRatio();
+			
+			bgfx::reset(width, height, BGFX_RESET_NONE);
+			_camera->SetProjectionMatrixPerspective(70.0f, aspect, 1.0f, 65536.0f);
```
bwrsandman, line 189: These magic numbers should be set as named constants and eventually should come from game data or config file.
I would recommend you make an entry in the Config struct and use it in all the places we set FOV and near and far.

bwrsandman (changes requested):

bwrsandman (approved):

src/Game.cpp
captnfab: I added three new parameters in Config struct. These are used at least at two point, at camera initialization, and on resize.
I used your static_cast presentation.

captnfab:
@bwrsandman : thanks for your reviews!
Is there some reference document for this project's coding style?

bwrsandman:
There isn't one yet. In general, I would say try to match the style of the files you're editing.
We have a clang-format formatting checker and config file.
We check for copyright also.
There's also a clang-tidy config but we don't yet have automatic checking for that #221.

One thing I often check for is that you follow the [google style order of header includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) and that headers do declarations and not implementations.

Special care for ECS code: Use Archtetypes instead of polymorphism (classes), Entities are plain old data and Systems do the logic.

bwrsandman (changes requested):
I tested it and it doesn't seem to work very well

Scaling up
Screenshot from 2022-04-19 15-43-14

Scaling Down
Screenshot from 2022-04-19 15-43-22


After a few resizes I get this followed by a crash:
Screenshot from 2022-04-19 15-43-35

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  bgfx: vcpkg/buildtrees/bgfx/src/5889d79ede-d8473b23d1.clean/src/renderer_vk.cpp:7830: FATAL (DebugCheck): vkQueueSubmit(m_queue, 1, &si, m_completedFence); VK error 0xfffffffc: VK_ERROR_DEVICE_LOST
[1]    482260 IOT instruction (core dumped)  ./cmake-build-presets/ninja-multi-vcpkg/bin/Debug/openblack -g 
```

captnfab:
Mmmm, there seems to be something weird happening when width is larger than 1280px.

captnfab:
PR still not ready for merge.

captnfab:
@bwrsandman : I haven't been able to reproduce your crash, but I fixed a couple of bugs that could have been responsible for it.

captnfab:
There's still one problem. To do the reset, I need to know the flags to apply.
I need to know if vsync needs to be used or not.

I'd say the vsync flag has to be saved somewhere in the Redrerer class, and the bgfx::reset needs to be moved to a Renderer::reset method.

raffclar:
> There's still one problem. To do the reset, I need to know the flags to apply. I need to know if vsync needs to be used or not.
> 
> I'd say the vsync flag has to be saved somewhere in the Redrerer class, and the bgfx::reset needs to be moved to a Renderer::reset method.

You can see the reset we use here:
https://github.com/openblack/openblack/blob/master/src/Renderer.cpp#L142
```c++
	init.resolution.reset = BGFX_RESET_NONE;
	if (vsync)
	{
		init.resolution.reset |= BGFX_RESET_VSYNC;
	}
```

Looks like we'll need to remember if vsync has been turned on or off in `Renderer`.

bwrsandman:
New changes have fixed it.

bwrsandman:
vsync comes from `args.vsync` in `Game`

src/Gui/Gui.cpp
```diff
@@ -209,6 +209,12 @@ bool Gui::ProcessEventSdl2(const SDL_Event& event)
 		io.KeySuper = ((SDL_GetModState() & KMOD_GUI) != 0);
 		return io.WantCaptureKeyboard;
 	}
+	case SDL_WINDOWEVENT:
+		if (event.window.event == SDL_WINDOWEVENT_SIZE_CHANGED)
+		{
+			bgfx::reset(static_cast<uint32_t>(event.window.data1), static_cast<uint32_t>(event.window.data2), BGFX_RESET_VSYNC);
```
bwrsandman, line 215: Why does this have to be caught in the Gui? This is a bad place to have this code.

bwrsandman: I've tested resizing without this and it doesn't contribute.

captnfab: In the event loop (https://github.com/openblack/openblack/blob/ddf190ae27a2779adb260c2719b87d09f8fbb0ce/src/Game.cpp#L131) the Gui is the first to handle SDL events.

By default, if an event is out of the window it is dropping it (https://github.com/openblack/openblack/blob/ddf190ae27a2779adb260c2719b87d09f8fbb0ce/src/Gui/Gui.cpp#L215)
The problem in this case is that the window cannot be scaled up.

So an important part in the event handling in Gui.cpp is the `return false`.

However, as you said, the bgfx::reset is useless. It's already present in the Game::ProcessEvents function. I'm removing it from the Gui.

captnfab: All that remains to do is store the reset flags in the renderer and reuse them during reset.

src/Gui/Gui.cpp
```diff
@@ -209,6 +209,11 @@ bool Gui::ProcessEventSdl2(const SDL_Event& event)
 		io.KeySuper = ((SDL_GetModState() & KMOD_GUI) != 0);
 		return io.WantCaptureKeyboard;
 	}
+	case SDL_WINDOWEVENT:
+		if (event.window.event == SDL_WINDOWEVENT_SIZE_CHANGED)
+		{
+			return false;
+		}
 	}
```
bwrsandman, line 217: Missing a `break `

captnfab: Fixed.

src/Gui/Gui.cpp
captnfab: I added a uint32_t _bgfxReset member in Renderer for saving the reset flags. This should solve the last issue we had.

bwrsandman:
LGTM. Just want to test it when I get home

bwrsandman:
It starts up like this without even resizing
Screenshot from 2022-04-20 20-53-37

captnfab:
> It starts up like this without even resizing. After resizing it's okay.
I can't reproduce this bug on my setups, but it must be related to the addition of `bgfx::reset(width, height, _bgfxReset);` in the `Renderer::ConfigureView` method.
I'll add a Renderer::Reset method and move the bgfx::reset code there, so that it's not called on init but only after resize.

raffclar:
> It starts up like this without even resizing. After resizing it's okay. Screenshot from 2022-04-20 20-53-37

Same issue but other than that works fine on Windows

captnfab:
@bwrsandman @raffclar : I believe the bug is now fixed.

I wasn't able to reproduce it, it might be related to the version of bgfx we are using.(I am using rev cae7c83f896d2b4740ce232f6224da16b949b224)

bwrsandman (approved):
Tested it and it works great

PR #374, opened 2022-04-19
logging: Undo regression to logger for debug builds

bwrsandman:
When we moved to static lib, we lost logging to the console for debug builds.


raffclar:
I'm able to see console logs again.

raffclar (approved):

PR #376, opened 2022-04-19
Clean up asset paths

raffclar:
Small clean up to improve consistency


raffclar:
Tested, it works. Packtool is fine too.

bwrsandman:
The other components are like this IIRC could you have a look

bwrsandman (approved):

raffclar:
> The other components are like this IIRC could you have a look

Created an issue
https://github.com/openblack/openblack/issues/377

PR #378, opened 2022-04-20
Use Loader API to load all raw textures

bwrsandman:
* Cleaned up our resource loading.
* If loading from a path, use the stem (base name without extension) as the id.
* Add viewer for textures

Screenshot from 2022-04-19 21-41-07

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -136,7 +136,6 @@ Game::Game(Arguments&& args)
 
 Game::~Game()
 {
-	_misc0aTexture.reset();
 	_water.reset();
 	_sky.reset();
 	_animationPack.reset();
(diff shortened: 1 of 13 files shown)


src/Game.cpp
```diff
@@ -436,17 +435,22 @@ bool Game::Run()
 	_sky = std::make_unique<Sky>();
 	_water = std::make_unique<Water>();
 
-	_misc0aTexture = std::make_unique<graphics::Texture2D>("misc0a.raw");
+	for (const auto& f : std::filesystem::directory_iterator {_fileSystem->GetGamePath() / _fileSystem->TexturePath()})
```
bwrsandman, line 438: Use `FindPath`

src/Game.cpp
```diff
@@ -436,17 +435,22 @@ bool Game::Run()
...
-		uint32_t size = width * height * bpp;
-		auto data = _fileSystem->ReadAll(_fileSystem->TexturePath() / _misc0aTexture->GetName());
-		assert(data.size() == size);
-		_misc0aTexture->Create(width, height, 1, graphics::Format::R8, graphics::Wrapping::ClampEdge, data.data(), data.size());
+		if (f.path().extension() == ".raw")
+		{
+			SPDLOG_LOGGER_DEBUG(spdlog::get("game"), "Loading raw texture: {}", f.path().stem().string());
+			try
+			{
+				textureManager.Load(("raw" / f.path().stem()).string(), f);
```
raffclar, line 445: The first argument to `Load` is `entt::id_type`. Each id should be unique

raffclar (changes requested):

src/Game.cpp
raffclar: I see now - I misread the arguments.

raffclar (approved):

Issue #384, opened 2022-04-21
Add an RNG manager


bwrsandman:
It would be nice to have a singleton manager for our random number generation needs.
- It should wrap std::uniform_int_distribution and a float distributor too
- It should be thread safe
- It should have an option to seed with a constant for testing but not during serious gameplay on release

ref: https://github.com/openblack/openblack/pull/285/files#diff-36ee7b5b5e3f725d1470d81e67fc2a906c396c41746e02886de29483aef0fa96R36-R39

PR #389, opened 2022-04-22
Rng manager

FalseIlyu:
Would this me in line with #384 ?

- [x] wrap std::uniform_int_distribution
- [x] wrap a float distributor
- [x] thread safe
- [x] Have an option to seed with a constant for testing

Maybe there should be a requirement for the result of the distribution to be the same accross platform.


src/Resources/RNGManager.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace openblack::ressources
+{
+class RNGManager
+{
+public:
+    static RNGManager& GetInstance() {
+        static RNGManager rngManager;
+        return rngManager;
+    };
```
bwrsandman, line 22: This needs to be in the cpp file

FalseIlyu:
This should make the RNG thread safe (no concurrent access to the underlying generator) and is (at least to me) the logical extension of using a shared generator but you're going to have a cumulative performance price as more object use it.

FalseIlyu:
I marked the task as completed as i believe the current implementation does the job. If we can agree on the general design and on the test [here](https://github.com/FalseIlyu/openblack/blob/rng_manager_ctest/test/rng_manager/test_rng_manager.cpp) i think we can mark this as ready for review.

```
Running main() from D:\openblack\vcpkg\buildtrees\gtest\src\ase-1.11.0-7d339f16d4.clean\googletest\src\gtest_main.cc
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from RngManager
[ RUN      ] RngManager.debug_predictable_mono
[       OK ] RngManager.debug_predictable_mono (380 ms)
[ RUN      ] RngManager.debug_predictable_multi
[       OK ] RngManager.debug_predictable_multi (1079 ms)
[----------] 2 tests from RngManager (1460 ms total)
[----------] Global test environment tear-down
[==========] 2 tests from 1 test suite ran. (1460 ms total)
[  PASSED  ] 2 tests.
```

bwrsandman:
General style note, we try to be consistent which means `camelCase` instead of `snake_case`.
Member variables have the underscore at the end, not the start.
Also clang-format doesn't catch this but try to add a new line at the end of each file.

bwrsandman:
I don't think resources is the right namespace, we just that namespace for resources like textures and meshes using entt's resources API.

src/Resources/RNGManager.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include <atomic>
```
bwrsandman, line 12: I don't think this include is necessary

src/Resources/RNGManager.h
```diff
@@ -0,0 +1,34 @@
...
+#include <mutex>
+#include <random>
+
+namespace openblack::resources
+{
+class RNGManager
+{
+public:
+	static RNGManager& GetInstance();
+	int NextInt(int min = 1, int max = 500);
```
bwrsandman, line 22: I think uint32 would be better and default value agent really useful for ints

bwrsandman (changes requested):

openblack.log
```diff
@@ -0,0 +1,990 @@
+[2022-04-23 21:42:33.566] [game] [info] [Game.cpp:101] current binary path: D:/openblack/cmake-build-presets/msvc-17-vcpkg/bin/Release
```
bwrsandman, line 1: You accidentally included the log file. This should be in gitignore

src/Resources/RNGManager.h
FalseIlyu: Should i keep allowing to define bounds on call of NextInt like in 7fca844?

bwrsandman:
I will test with my branch and get back to you. It might take a few days.

src/Common/RNGManager.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+RNGManager& RNGManager::GetInstance()
+{
+	static RNGManager rngManager;
+	return rngManager;
+}
+
+uint32_t RNGManager::NextInt(uint32_t min, uint32_t max)
+{
+	static std::uniform_int_distribution<uint32_t> int_dist(min, max);
```
bwrsandman, line 22: This is not thread safe and it doesn't make sense to have it thread local either.

src/Common/RNGManager.h
```diff
@@ -0,0 +1,33 @@
...
+#include <mutex>
+#include <random>
+
+namespace openblack
+{
+class RNGManager
+{
+public:
+	static RNGManager& GetInstance();
+	uint32_t NextInt(uint32_t min, uint32_t max);
```
bwrsandman, line 21: NextUint32

src/Common/RNGManager.cpp
```diff
@@ -0,0 +1,50 @@
...
+	return float_dist(generator);
+}
+
+bool RNGManager::SetDebugMode(bool isDebug, int seed)
+{
+	std::lock_guard<std::mutex> safe_lock(_generatorLock);
+	_debugRng = true;
+	_generator.seed(seed);
+	return _debugRng == isDebug;
+}
```
bwrsandman, line 50: Line return

src/Common/RNGManager.cpp
```diff
@@ -0,0 +1,50 @@
...
+		std::lock_guard<std::mutex> safe_lock(_generatorLock);
+		return int_dist(_generator);
+	}
+	thread_local std::mt19937 generator;
+	return int_dist(generator);
+}
+
+float RNGManager::NextFloat(float min, float max)
+{
+	static std::uniform_real_distribution<float> float_dist(min, max);
```
bwrsandman, line 34: Not thread safe.

src/Common/RNGManager.h
```diff
@@ -0,0 +1,33 @@
...
+	float NextFloat(float min = 0.0, float max = 1.0);
+	bool SetDebugMode(bool is_debug, int seed);
+
+private:
+	std::mt19937 _generator;
+	std::mutex _generatorLock;
+	bool _debugRng = false;
+	RNGManager() {};
```
bwrsandman, line 29: The definition should be in the implementation file (*.cpp)

src/Common/RNGManager.h
```diff
@@ -0,0 +1,33 @@
...
+	RNGManager(const RNGManager&) = delete;
+	RNGManager& operator=(const RNGManager&) = delete;
```
bwrsandman, line 31: Deletes should have public visibility

src/Common/RNGManager.h
```diff
@@ -0,0 +1,33 @@
...
+public:
+	static RNGManager& GetInstance();
+	uint32_t NextInt(uint32_t min, uint32_t max);
+	float NextFloat(float min = 0.0, float max = 1.0);
+	bool SetDebugMode(bool is_debug, int seed);
+
+private:
+	std::mt19937 _generator;
+	std::mutex _generatorLock;
+	bool _debugRng = false;
```
bwrsandman, line 28: Please define in the constructor you have

src/Common/RNGManager.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <mutex>
+#include <random>
+
+namespace openblack
+{
+class RNGManager
+{
+public:
+	static RNGManager& GetInstance();
```
bwrsandman, line 20: In order to be consistent please name this `instance()`

src/Common/RNGManager.h
```diff
@@ -0,0 +1,33 @@
...
+	uint32_t NextInt(uint32_t min, uint32_t max);
+	float NextFloat(float min = 0.0, float max = 1.0);
```
bwrsandman, line 22: Not really useful to have default values.

bwrsandman (changes requested):

bwrsandman:
I think you're better off merging NextInt and NextFloat into a template.
Since you're already including `mutex` and `random` in the header, I'm not worried about added include leaks

.gitignore
```diff
@@ -2,6 +2,9 @@
 .idea/
 cmake-build-*/
 
+#Log files
+*.log
+
```
bwrsandman, line 7: You can remove this change now

src/Common/RNGManager.h
```diff
@@ -0,0 +1,45 @@
...
+		    std::conditional_t<std::is_integral_v<T>, std::uniform_int_distribution<T>, std::uniform_real_distribution<T>>;
+		dist_t dist(min, max);
+		if (_debugRng)
+		{
+			std::lock_guard<std::mutex> safe_lock(_generatorLock);
+			return dist(_generator);
+		}
+		thread_local std::mt19937 generator;
+		return dist(generator);
+	};
```
bwrsandman, line 34: Extra `;`

bwrsandman:
I've tested it and it works really well and exactly how I imagined. Good work!

bwrsandman (approved):

PR #390, opened 2022-04-22
Use locator for exisiting raw files

### LINKED ISSUE #388: Use resource locator for raw textures to find previouly used raw textures
The following raw textures are loaded differently than misc0a.raw
* Sky.raw
* smallbumpa.raw

They should be available as the following
* raw/Sky
* raw/smallbumpa

bwrsandman:
Also added Skya to the water shader (diffuse is confusingly called Sky for water) for alpha but with alpha blending off, it has no effect 
Resolves #388

### DIFF
--- a/src/3D/LandIsland.cpp
+++ b/src/3D/LandIsland.cpp
@@ -10,7 +10,6 @@
 #include "LandIsland.h"
 
 #define STB_IMAGE_WRITE_IMPLEMENTATION
-#include "Common/FileSystem.h"
 #include "Common/IStream.h"
 #include "Common/stb_image_write.h"
 #include "Dynamics/LandBlockBulletMeshInterface.h"
@@ -31,14 +30,6 @@ const float LandIsland::CellSize = 10.0f;
 LandIsland::LandIsland()
     : _blockIndexLookup {0}
 {
-	auto& filesystem = Game::instance()->GetFileSystem();
-	auto file = filesystem.Open(filesystem.TexturePath() / "smallbumpa.raw", FileMode::Read);
-	auto* smallbumpa = new uint8_t[file->Size()];
-	file->Read(smallbumpa, file->Size());
-
-	_textureSmallBump = std::make_unique<Texture2D>("LandIslandSmallBump");
-	_textureSmallBump->Create(256, 256, 1, Format::R8, Wrapping::Repeat, smallbumpa, static_cast<uint32_t>(file->Size()));
-	delete[] smallbumpa;
 }
 
 LandIsland::~LandIsland() = default;
--- a/src/3D/Water.cpp
+++ b/src/3D/Water.cpp
@@ -28,16 +28,6 @@ Water::Water()
 	_reflectionFrameBuffer =
 	    std::make_unique<FrameBuffer>("Reflection", static_cast<uint16_t>(1024), static_cast<uint16_t>(1024),
 	                                  graphics::Format::RGBA8, graphics::Format::Depth24Stencil8);
-
-	// water texture (256x256 RAW RGB 24bpp)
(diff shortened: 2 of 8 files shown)


src/Renderer.cpp
```diff
@@ -443,9 +448,12 @@ void Renderer::DrawPass(const DrawSceneDesc& desc) const
 		{
 			for (auto& block : desc.island.GetBlocks())
 			{
+				static const auto textureId = entt::hashed_string("raw/smallbumpa");
```
bwrsandman, line 451: This constant should be on the class

src/Renderer.cpp
```diff
@@ -428,8 +428,13 @@ void Renderer::DrawPass(const DrawSceneDesc& desc) const
 			mesh->GetIndexBuffer().Bind(mesh->GetIndexBuffer().GetCount(), 0);
 			mesh->GetVertexBuffer().Bind();
 			bgfx::setState(BGFX_STATE_DEFAULT);
-			waterShader->SetTextureSampler("s_diffuse", 0, *desc.water._texture);
-			waterShader->SetTextureSampler("s_reflection", 1, desc.water.GetFrameBuffer().GetColorAttachment());
+			static const auto diffuseId = entt::hashed_string("raw/Sky");
+			static const auto alphaId = entt::hashed_string("raw/Skya");
```
bwrsandman, line 432: These constants should be on the class

src/3D/LandIsland.cpp
```diff
@@ -31,14 +30,6 @@ const float LandIsland::CellSize = 10.0f;
 LandIsland::LandIsland()
     : _blockIndexLookup {0}
 {
-	auto& filesystem = Game::instance()->GetFileSystem();
-	auto file = filesystem.Open(filesystem.TexturePath() / "smallbumpa.raw", FileMode::Read);
-	auto* smallbumpa = new uint8_t[file->Size()];
-	file->Read(smallbumpa, file->Size());
-
-	_textureSmallBump = std::make_unique<Texture2D>("LandIslandSmallBump");
-	_textureSmallBump->Create(256, 256, 1, Format::R8, Wrapping::Repeat, smallbumpa, static_cast<uint32_t>(file->Size()));
```
raffclar, line 40: Getting odd visual artifacts on my end. Different format perhaps?

image

raffclar: Logic within the loader for comparison:

```texture->Create(width, height, 1, format, graphics::Wrapping::ClampEdge, data.data(), static_cast<uint32_t>(data.size()));```

File: src/3D/Water.cpp [hunk 4]
```diff
@@ -28,16 +28,6 @@ Water::Water()
...
 	    std::make_unique<FrameBuffer>("Reflection", static_cast<uint16_t>(1024), static_cast<uint16_t>(1024),
 	                                  graphics::Format::RGBA8, graphics::Format::Depth24Stencil8);
-
-	// water texture (256x256 RAW RGB 24bpp)
-	auto& filesystem = Game::instance()->GetFileSystem();
-	auto const& waterTextureData = filesystem.ReadAll(filesystem.TexturePath() / "Sky.raw");
-	const uint16_t textureWidth = 256, textureHeight = 256;
-
-	_texture = std::make_unique<Texture2D>("Water");
-	_texture->Create(textureWidth, textureHeight, 1, Format::RGB8, Wrapping::Repeat, waterTextureData.data(),
```
raffclar, line 38: Same issue here. The water plane has no texture and the above differs from the texture loader

image

raffclar (changes requested):

raffclar (approved):
Nice one 👍

Issue #123, opened 2019-10-16
[valgrind] Memory leak in `openblack::LHScriptX::ScriptCommandParameter::_value`


bwrsandman:
```
1,747 (1,344 direct, 403 indirect) bytes in 42 blocks are definitely lost in loss record 318 of 345
operator new(unsigned long)
ScriptCommandParameter
GetParameter(openblack::LHScriptX::Token&)
openblack::LHScriptX::Script::runCommand(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&, std::vector<openblack::LHScriptX::Token, std::allocator> const&)
openblack::LHScriptX::Script::Load(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::LoadMap(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::Run()
main

---

80,038 (79,232 direct, 806 indirect) bytes in 2,476 blocks are definitely lost in loss record 344 of 345
operator new(unsigned long)
ScriptCommandParameter
GetParameter(openblack::LHScriptX::Token&)
openblack::LHScriptX::Script::runCommand(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&, std::vector<openblack::LHScriptX::Token, std::allocator> const&)
openblack::LHScriptX::Script::Load(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::LoadMap(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::Run()
main
```

Memory allocated in https://github.com/openblack/openblack/blob/3409c64a4a2be63edb430a9887ef68f69dab7da9/src/LHScriptX/CommandSignature.h#L51-L52 is never freed.

PR #391, opened 2022-04-22
Memory leaks in LHScriptX::Token::_u and  CommandSignature

### LINKED ISSUE #122: [valgrind] Memory leak in `openblack::LHScriptX::Token::_u`
```
1,617 (1,344 direct, 273 indirect) bytes in 42 blocks are definitely lost in loss record 317 of 345
operator new(unsigned long)
MakeIdentifierToken
openblack::LHScriptX::Lexer::gatherIdentifer()
openblack::LHScriptX::Lexer::GetToken()
advanceToken
openblack::LHScriptX::Script::Load(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::LoadMap(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::Run()
main

---

32 bytes in 1 blocks are definitely lost in loss record 74 of 345
operator new(unsigned long)
MakeIdentifierToken
openblack::LHScriptX::Lexer::gatherIdentifer()
openblack::LHScriptX::Lexer::GetToken()
peekToken
openblack::LHScriptX::Script::Load(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::LoadMap(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::Run()
main

---

61,284 (60,768 direct, 516 indirect) bytes in 1,899 blocks are definitely lost in loss record 342 of 345
operator new(unsigned long)
MakeStringToken
openblack::LHScriptX::Lexer::gatherString()
openblack::LHScriptX::Lexer::GetToken()
advanceToken
openblack::LHScriptX::Script::Load(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::LoadMap(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::Run()
main

---

92,394 (74,912 direct, 17,482 indirect) bytes in 2,341 blocks are definitely lost in loss record 345 of 345
operator new(unsigned long)
MakeIdentifierToken
openblack::LHScriptX::Lexer::gatherIdentifer()
openblack::LHScriptX::Lexer::GetToken()
advanceToken
openblack::LHScriptX::Script::Load(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::LoadMap(std::__cxx11::basic_string<char, std::char_traits, std::allocator> const&)
openblack::Game::Run()
main
```

Memory allocated in
https://github.com/openblack/openblack/blob/3409c64a4a2be63edb430a9887ef68f69dab7da9/src/LHScriptX/Lexer.h#L80
https://github.com/openblack/openblack/blob/3409c64a4a2be63edb430a9887ef68f69dab7da9/src/LHScriptX/Lexer.h#L86
 is never freed.

captnfab:
Fixes #122, #123

Remove use of new (without delete) in Token creation.
Instead, a plain std::string is present in Token, next to the Union.
Overhead (once during script loading):
- One string copy per command
- One std::string init/deletion per numerical or operator token

### DIFF
--- a/src/LHScriptX/CommandSignature.h
+++ b/src/LHScriptX/CommandSignature.h
@@ -37,12 +37,10 @@ public:
 	    : _type(type)
 	{
 	}
-	ScriptCommandParameter(const std::string value)
+	ScriptCommandParameter(const std::string& value)
 	    : _type(ParameterType::String)
 	{
-		// TODO: Hacky, avoid new
-		_value._string = new std::string();
-		*_value._string = value;
+		_string = value;
 	}
 	ScriptCommandParameter(float value)
 	    : _type(ParameterType::Float)
@@ -58,12 +56,12 @@ public:
 
 	[[nodiscard]] ParameterType GetType() const { return _type; };
 
-	[[nodiscard]] std::string& GetString() const { return *_value._string; }
+	[[nodiscard]] const std::string& GetString() const { return _string; }
 	[[nodiscard]] float GetFloat() const { return _value._float; }
 	[[nodiscard]] int32_t GetNumber() const { return _value._number; }
 	void GetVector(float& x, float& y, float& z) const;
 
-	void SetString(const std::string& value) { *_value._string = value; }
+	void SetString(const std::string& value) { _string = value; }
 	void SetFloat(float value) { _value._float = value; };
 	void SetNumber(int32_t value) { _value._number = value; }
 	void SetVector(float x, float y, float z)
@@ -78,11 +76,11 @@ private:
 
 	union
 	{
-		std::string* _string;
 		float _float;
 		int32_t _number;
 		float _vector[3];
 	} _value;
+	std::string _string;
 };
 
 typedef std::vector<ScriptCommandParameter> ScriptCommandParameters;
--- a/src/LHScriptX/Lexer.h
+++ b/src/LHScriptX/Lexer.h
@@ -68,13 +68,13 @@ public:
 	static Token MakeIdentifierToken(const std::string& value)
 	{
 		Token tok(Type::Identifier);
-		tok.u_.identifierValue = new std::string(value);
+		tok.s_ = value;
 		return tok;
 	}
 	static Token MakeStringToken(const std::string& value)
 	{
 		Token tok(Type::String);
-		tok.u_.stringValue = new std::string(value);
+		tok.s_ = value;
 		return tok;
 	}
 	static Token MakeIntegerToken(int value)
@@ -103,8 +103,8 @@ public:
 	[[nodiscard]] bool IsOP(Operator op) const { return this->type_ == Type::Operator && this->u_.op == op; }
 
 	// todo: assert check the type for each of these?
-	[[nodiscard]] const std::string& Identifier() const { return *this->u_.identifierValue; }
-	[[nodiscard]] const std::string& StringValue() const { return *this->u_.stringValue; }
+	[[nodiscard]] const std::string& StringValue() const { return this->s_; }
+	[[nodiscard]] const std::string& Identifier() const { return this->s_; }
 	[[nodiscard]] const int* IntegerValue() const { return &this->u_.integerValue; }
 	[[nodiscard]] const float* FloatValue() const { return &this->u_.floatValue; }
 	[[nodiscard]] Operator Op() const { return this->u_.op; }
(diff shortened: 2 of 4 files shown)


bwrsandman:
Is it the same memory leak as this?
https://github.com/openblack/openblack/runs/6134871643?check_suite_focus=true

bwrsandman:
There seems to also be a leak here https://github.com/openblack/openblack/blob/master/src/LHScriptX/CommandSignature.h#L44

captnfab:
> There seems to also be a leak here https://github.com/openblack/openblack/blob/master/src/LHScriptX/CommandSignature.h#L44

Yep, this is #123.

> Is it the same memory leak as this? https://github.com/openblack/openblack/runs/6134871643?check_suite_focus=true

Yep, as for now, this MR addresses records 4 and 5. The other leak you mentioned corresponds to record 6.

I'll take a look at it to address it on this MR.

bwrsandman:
Something just occurred to me. Is it possible to avoid a string copy by using a string view?

bwrsandman:
I checked with valgrind, all memory leaks are fixed.

captnfab:
> Something just occurred to me. Is it possible to avoid a string copy by using a string view?

I didn't know about `string_view`. Indeed, it seems we could benefit from massively using string views everywhere we got `const std::string`.

I changed strings to string views in the Lexer and Token. However, changing from strings to string views in ScriptCommandParameter would require major changes since script functions, and functions called from them, often use strings where they could use string_views.

captnfab:
Is the macOs builder using an old/non-standard version of C++?

src/LHScriptX/Lexer.h
```diff
@@ -121,12 +122,11 @@ class Token
 	Type type_;
 	union
 	{
-		std::string* identifierValue;
-		std::string* stringValue;
 		int integerValue;
 		float floatValue;
 		Operator op;
 	} u_;
+	std::string_view s_;
```
handsomematt, line 129: This really doesn't seem right, should it not be a std::string

src/LHScriptX/CommandSignature.h
```diff
@@ -37,12 +38,10 @@ class ScriptCommandParameter
 	    : _type(type)
 	{
 	}
-	ScriptCommandParameter(const std::string value)
+	ScriptCommandParameter(const std::string_view& value)
```
handsomematt, line 41: Pass string_view by value and without const

src/LHScriptX/Lexer.h
```diff
@@ -103,8 +104,8 @@ class Token
 	[[nodiscard]] bool IsOP(Operator op) const { return this->type_ == Type::Operator && this->u_.op == op; }
 
 	// todo: assert check the type for each of these?
-	[[nodiscard]] const std::string& Identifier() const { return *this->u_.identifierValue; }
-	[[nodiscard]] const std::string& StringValue() const { return *this->u_.stringValue; }
+	[[nodiscard]] const std::string_view StringValue() const { return this->s_; }
```
handsomematt, line 107: Returning string_view doesn't seem right, they're just pointers being flung around.

src/LHScriptX/Lexer.h
handsomematt: Who owns the string it's viewing? The Lexer? If it's intended and known let's stick a comment on it saying exactly what it's viewing.

handsomematt (changes requested):

src/LHScriptX/Lexer.h
captnfab: Oh, I thought the string_view were a bit like shared_ptr, with the string allocated in the heap. But it actually is more like a raw pointer which could become dangling if the source string is missing?

In this case, maybe I should stick to the first two commits of this MR and forget about the two lasts? We'll have more string copies, but string ownership mill be clearer.

captnfab:
Version with string_views has been moved there: https://github.com/captnfab/openblack/tree/issue-122-with-stringviews
I'm pushing back version with plain strings.

src/LHScriptX/Lexer.h
captnfab: Reverted back to string.

src/LHScriptX/CommandSignature.h
captnfab: Reverted to string.

src/LHScriptX/Lexer.h
captnfab: Reverted to string.

bwrsandman:
Can you rebase on master?

captnfab:
rebased on 95597fae77f312e1b0b80834fd4ec262f2c7b8dd since CI on master fails because of missing EOL at EOF.

handsomematt (approved):

PR #387, opened 2022-04-21
Activate CTests using GTests

bwrsandman:
Resolves #385 

A few bugs also popped up:
* On metal, the renderer is running in single-thread mode. But when the renderer restarts, it is in multithreaded mode.
* l3d's write was accessing invalid memory
* Camera bookmark was not checking if the resources were there.

PR #393, opened 2022-04-23
Load scene tests

bwrsandman:
Add some unit tests to catch errors in scene file loading. This is useful to build on for more advanced tests like those in #338 

The test revealed some undefined behaviour when trying to get island before one is loaded.

- [x] Depends on #387


captnfab:
@bwrsandman: do you plan to move you bug fixes out of this PR so that they can be more easily reviewed and merged?

PR #394, opened 2022-04-23
Use stringviews to share strings between Lexer and Tokens

captnfab:
Based on https://github.com/openblack/openblack/pull/391 but uses stringviews to share strings between Lexer and Tokens.

### DIFF
--- a/src/LHScriptX/Lexer.cpp
+++ b/src/LHScriptX/Lexer.cpp
@@ -184,7 +184,9 @@ Token Lexer::gatherIdentifer()
 		current_++;
 	}
 
-	return Token::MakeIdentifierToken(std::string_view(id_start, current_));
+	// return Token::MakeIdentifierToken(std::string_view(id_start, current_));
+	// Temporary fix for AppleClang 13.0.0.13000029 from XCode 13.2.1
+	return Token::MakeIdentifierToken(std::string_view(&(*id_start), std::distance(id_start, current_)));
 }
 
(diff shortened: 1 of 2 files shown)


captnfab:
And there's this issue on macOs, I don't know why it doesn't know the constructor `std::string_view(iterator_start, iterator_end)`…

bwrsandman:
According to the build, it's AppleClang 13.0.0.13000029 from XCode 13.2.1

bwrsandman:
I feel like I had a similar issue with spans here https://github.com/openblack/openblack/pull/336

bwrsandman:
Maybe this constructor will work
```c++
constexpr basic_string_view( const CharT* s, size_type count );
```

bwrsandman:
Or we can try setting clang 13.0.1 with `$(brew --prefix llvm)/bin/clang`

handsomematt:
Does this actually work? The Lexer is scoped to Script::load, making it a string_view keeps the ownership on the Lexer which would get deleted?

captnfab:
@bwrsandman: yeah, correct. Given how the tokens are used now, they cannot outlive the Lexer, so it's fine. I agree with you about `source_` not being const.

I also agree about the fact that this would bring only little improvement. It was just a pretense to study the lexer and play with stringviews. So if you don't see this being merged, I don't mind closing the PR.

bwrsandman:
You can close it

PR #401, opened 2022-04-27
actions: Add check for EOL at EOF

bwrsandman:
We shouldn't have to police everyone to check their line endings at the end of files.

https://github.com/marketplace/actions/linelint


raffclar (approved):
Your PR is failing because of the new check 👍

raffclar:
And... you just pushed a commit to fix the new check :sweat_smile:

bwrsandman:
@handsomematt review dog complains:

> reviewdog: This GitHub token doesn't have write permission of Review API [1], 
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
[1]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target, 
[2]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#logging-commands


https://github.com/openblack/openblack/runs/6186963907?check_suite_focus=true#step:8:182

handsomematt:
Everything looks good here, regarding the permissions do we not need to simply add `REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}` ?

bwrsandman:
Normally, action suggester is using `${{ github.token }}` by default

Issue #402, opened 2022-04-27
Automatically build android apk


bwrsandman:
We have android github actions to build android libraries but we do not run `gradle assemble`

This is a more complex case because it runs all configurations and stores them in one artifact.

We could make a dependent job which builds on top of the `Cross Compile/build` android-armeabi-v7a-vcpkg, android-arm64-v8a-vcpkg, android-x86-vcpkg, android-x86_64-vcpkg presets and somehow re-use the configuration and artifacts.

Another option would be to just create one job that runs `gradle assemble`.

PR #403, opened 2022-04-27
Android: Lint and build APKs

bwrsandman:
Fixes and tests building the APK on windows.
Also fixes gradle lint and the apk name
Fixes android failing to find main function

You can now get the openblack apk from the workflow artifacts. E.g. https://github.com/openblack/openblack/suites/6326800004/artifacts/227483860

Implements #402

TODO:
- [x] Use https://github.com/marketplace/actions/gradle-build-action 
- [x] Depends on #417
- [ ] Find a way to cache config step in apk build because it currently takes > 30 minutes on two runners.

### DIFF
--- a/.github/workflows/ci-cross-compile.yml
+++ b/.github/workflows/ci-cross-compile.yml
@@ -124,12 +124,12 @@ jobs:
       - name: Run Gradle Configure CMake (Debug)
         uses: gradle/gradle-build-action@v2
         with:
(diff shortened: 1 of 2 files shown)


bwrsandman:
I think I traced the source of the error in gradle from window to the linking with SDL.
```
C/C++: ld: error: undefined symbol: __fread_chk 
```

PR #406, opened 2022-04-27
Use path instead of string in components and apps

### LINKED ISSUE #377: Refactor asset/resource classes to use std::filesystem where possible
We should stop std::string as a way to represent filesystem paths. We now have C++20 and it has std::filesystem

captnfab:
Fixes #377 

In components:

- [x] anm
- [x] l3d
- [x] morph
- [x] lnd
- [x] pack (fix)

In apps:

- [x] anmtool
- [x] l3dtool
- [x] morphtool
- [x] lnd
- [x] packtool

### DIFF
--- a/apps/anmtool/anmtool.cpp
+++ b/apps/anmtool/anmtool.cpp
@@ -94,12 +94,12 @@ struct Arguments
 	Mode mode;
 	struct Read
 	{
-		std::vector<std::string> filenames;
+		std::vector<std::filesystem::path> filenames;
 	} read;
 	struct Write
 	{
-		std::string outFilename;
-		std::string gltfFile;
+		std::filesystem::path outFilename;
+		std::filesystem::path gltfFile;
 	} write;
 };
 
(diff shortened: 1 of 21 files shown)


captnfab:
It looks to me that every time the GetFilename method is used, the aim is to get a string and print it.
So maybe GetFilename should be return `std::string _filename.string()` instead of `std::filesystem::path& _filename`

apps/anmtool/anmtool.cpp
```diff
@@ -133,20 +133,20 @@ bool parseOptions(int argc, char** argv, Arguments& args, int& return_code)
...
+  options.positional_help("[read|write] [OPTION...]");
+  options.add_options()
+    ("H,header", "Print Header Contents.", cxxopts::value<std::vector<std::filesystem::path>>())
+    ("l,list-keyframes", "List Keyframes.", cxxopts::value<std::vector<std::filesystem::path>>())
+    ("k,keyframe-content", "View Keyframe Contents", cxxopts::value<std::vector<std::filesystem::path>>())
+    ;
+  options.add_options("write from and to glTF format")
+    ("o,output", "Output file (required).", cxxopts::value<std::filesystem::path>())
+    ("i,input-mesh", "Input file (required).", cxxopts::value<std::filesystem::path>())
+    ;
```
bwrsandman, line 149: Hard to tell on Github but are these now tabs?

bwrsandman (approved):

apps/anmtool/anmtool.cpp
captnfab: No, these are still spaces.
I didn't pay attention to that, but it looks like since this section has clang-format disabled, my editor automatically used its default indentation configuration, that is "two spaces"-indentation.

PR #407, opened 2022-04-27
RNGManager access through Locator

FalseIlyu:
Simplest way to make the RNGManager access through Locator

### DIFF
--- a/src/Common/RNG.h
+++ b/src/Common/RNG.h
@@ -31,6 +31,8 @@ public:
 		}
 		return dist(generator());
 	}
+	virtual ~RNGInterface() {}
+
 protected:
 	virtual std::mt19937& generator() = 0;
 	virtual std::mutex& lockAll() = 0;
--- a/src/Common/RNG.h
+++ b/src/Common/RNG.h
@@ -31,7 +31,7 @@ public:
 		}
 		return dist(generator());
 	}
-	virtual ~RNGInterface() {}
+	virtual ~RNGInterface() = default;
 
 protected:
 	virtual std::mt19937& generator() = 0;
--- a/src/Common/RNG.h
+++ b/src/Common/RNG.h
@@ -17,7 +17,7 @@ namespace openblack
 class RNGInterface
 {
 public:
-	virtual bool SetDebugMode(bool is_debug, int seed) = 0;
+	virtual bool SetDebugMode(bool isDebug, int seed) = 0;
 	template <typename T, typename E = typename std::enable_if<std::is_arithmetic_v<T>>::type>
 	T NextValue(T min, T max)
 	{
--- a/src/Common/RNG.h
+++ b/src/Common/RNG.h
@@ -1,0 +1,28 @@
+/*******************************************************************************
+ * Copyright (c) 2018-2022 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ ******************************************************************************/
+
+#include "RNGManager.h"
+
+namespace openblack
+{
+class RNGInterface
+{
+public:
+    virtual RNGManager& getRNG() = 0;
+};
+
+class RNG final: public RNGInterface
+{
+public:
+    RNGManager& getRNG() override { return _rngManager; }
+private:
+    RNGManager _rngManager;
+};
+
+} // namespace openblack
--- a/src/Common/RNG.h
+++ b/src/Common/RNG.h
@@ -17,14 +17,14 @@ namespace openblack
 class RNGInterface
 {
 public:
-	virtual bool SetDebugMode(bool isDebug, int seed) = 0;
+	virtual bool SetSeed(int seed) = 0;
 	template <typename T, typename E = typename std::enable_if<std::is_arithmetic_v<T>>::type>
 	T NextValue(T min, T max)
 	{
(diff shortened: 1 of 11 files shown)


src/Locator.h
```diff
@@ -19,5 +20,6 @@ namespace openblack
 struct Locator
 {
 	using resources = entt::service_locator<resources::ResourcesInterface>;
+	using rng = entt::service_locator<RNGManager>;
```
raffclar, line 23: Create an interface for RNGManager called `RNGManagerInterface`. We don't want to couple the locator to an implementation.

src/Game.cpp
```diff
@@ -418,6 +418,7 @@ bool Game::Update()
 bool Game::Initialize()
 {
 	Locator::resources::set<resources::Resources>();
+	Locator::rng::set<RNGManager>();
```
raffclar, line 421: 👍

raffclar (changes requested):

src/Common/RNG.h
```diff
@@ -0,0 +1,41 @@
...
+		    std::conditional_t<std::is_integral_v<T>, std::uniform_int_distribution<T>, std::uniform_real_distribution<T>>;
+		dist_t dist(min, max);
+		if (isDebug())
+		{
+			std::lock_guard<std::mutex> safe_lock(lockAll());
+			return dist(generator());
+		}
+		return dist(generator());
+	}
+	virtual ~RNGInterface() {}
```
bwrsandman, line 34: Move the implementation to the implementation file.
```
~RNGInterface::RNGInterface() = default;
```

src/Common/RNGManager.cpp
```diff
@@ -16,16 +17,30 @@ RNGManager::RNGManager()
...
 	return _debugRng == isDebug;
 }
+
+std::mt19937& RNGManager::generator()
+{
+	if (isDebug())
+	{
+		return _generator;
+	}
+	thread_local std::mt19937 tgenerator(time(nullptr));
```
bwrsandman, line 34: Fix these warning:
>  'argument': conversion from 'time_t' to 'unsigned int', possible loss of data

bwrsandman (changes requested):
I feel like the design here is wrong.

If we're going through the Locator pattern, debug and non-debug should be two different implementations.
of `RNGInterface`.

bwrsandman (changes requested):
Same as previously. With the locator pattern, it's best to have two implementations of RNG.h, which the name should be improved like: RandomNumberManager.h
It should provide two interfaces to the locator:
* RandomNumberManagerProduction.h
* RandomNumberManagerTesting.h

src/Game.cpp
```diff
@@ -423,6 +424,7 @@ bool Game::Update()
...
```
bwrsandman, line 427: This is where you would either set the debug or production rng implementation

src/Common/RandomNumberManagerProduction.cpp
```diff
@@ -0,0 +1,36 @@
...
+std::mt19937& RandomNumberManagerProduction::generator()
+{
+	thread_local std::mt19937 tGenerator(static_cast<unsigned int>(time(nullptr)));
+	return tGenerator;
+}
+
+std::mutex& RandomNumberManagerProduction::lockAll()
+{
+	std::mutex rMutex;
+	return rMutex;
```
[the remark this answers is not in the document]
bwrsandman: Is a mutex necessary in production?

src/Common/RandomNumberManagerProduction.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "RandomNumberManagerProduction.h"
+#include <ctime>
+
+using namespace openblack;
+
+bool RandomNumberManagerProduction::SetSeed(int seed)
+{
+	generator().seed(seed);
+	return true;
+}
```
bwrsandman, line 19: Do we need to manually set seed in production?

bwrsandman (changes requested):

src/Common/RandomNumberManagerProduction.cpp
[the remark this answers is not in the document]
FalseIlyu: We need to have the lock callable for NextValue even if it is not used, it shouldn't be necessary for production but i don't quite know what could be done to have NextValue() be separate for test and production as most of the idea i found need for NextValue to not use template.

src/Common/RandomNumberManagerProduction.cpp
bwrsandman: In that case wouldn't that be another implementation?
Maybe add a MultiplayerSynced implementation

src/Common/RandomNumberManager.h
```diff
@@ -1,45 +1,43 @@
...
  * For a complete list of all authors, please refer to contributors.md
  * Interested in contributing? Visit https://github.com/openblack/openblack
  *
  * openblack is licensed under the GNU General Public License version 3.
- *****************************************************************************/
+ ******************************************************************************/
 
 #pragma once
 
+#include <functional>
```
bwrsandman, line 12: Are you sure you need this include?

bwrsandman (approved):
My issue remains that it would be ideal if we could have the RNG manager interface not require optional or mutex.

bwrsandman (changes requested):
Please fix the warnings in `Files changed`

raffclar (approved):

bwrsandman (approved):

Issue #410, opened 2022-04-29
Merge debug and release vcpkg tests


bwrsandman:
Vcpkg config takes the longest time in our CI.
It builds both debug and release for dependencies.
Having a permutation for debug and release doubles the config time if there's a cache miss and doubles the space needed for cache hits.

PR #412, opened 2022-04-29
actions: Merge config permutations

bwrsandman:
Implements #410 

Since the configure step using vcpkg builds both `dbg` and `rel`, we're losing a lot of time to rebuilding vcpkg dependencies by doing config in the two types of config builds.
In the case of a cache miss, it can take 15-25 minutes to rebuild the vcpkg dependencies for both Debug and Release builds.
Merging both debug and release into the same build allows us to remove one Config step (15-25 minutes).
It also cuts the number of builds by half which would reduce the load on the caching which probably will result in more cache hits.
Building openblack after configuration takes about 2-4 minutes so in the case of a cache hit, the build takes 2-4 minutes longer. However, there are less builds running per commit, therefore more runners can go simultaneously.

### DIFF
--- a/.github/workflows/ci-cross-compile.yml
+++ b/.github/workflows/ci-cross-compile.yml
@@ -50,12 +50,14 @@ jobs:
         with:
           vcpkgJsonGlob: 'vcpkg.json'
 
-      - name: Run CMake+vcpkg to generate (Config).
+      - name: Configure CMake+vcpkg.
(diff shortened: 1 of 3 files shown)


bwrsandman:
Weirdly, the windows builds suddenly forget that they're in C++20

handsomematt (approved):

PR #413, opened 2022-05-01
mock: Use InfoConstants.h to generate mock info.dat

bwrsandman:
Using c++ to generate info.dat make more sense since the header can be used to find offsets directly.

I thought about adding this to apps but it seems too mock specific. Maybe in the future it would make sense to put it there.

The app will be useful for adding mesh data for the mobile wall hug tests


test/mock/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 find_package (Python3 COMPONENTS Interpreter)
 
+add_executable(gen_info gen_info.cpp)
+target_include_directories(gen_info PRIVATE ../../src)
+target_link_libraries(gen_info PRIVATE glm::glm)
```
bwrsandman, line 5: Should have cxxoopts too

PR #422, opened 2022-05-17
Script: Parse Vector arguments as vectors

bwrsandman:
An exception was being created but never thrown and the vector param was never implemented.
Instead, the shim layer was converting strings to vectors.


handsomematt (approved):
This is good, only part I'm iffy on is checking if it's a vector based on the string, is there really nothing that says "this argument is a vector!"

bwrsandman:
Nothing from the line other than the string contents.
The function expects a vector which is how we got away with it previously

PR #425, opened 2022-05-18
ci: Remove ZedThree/clang-tidy-review as it's unused

bwrsandman:


bwrsandman:
- [ ] Check that existing warnings are still present

Issue #427, opened 2022-05-22
Failed assert 'Array subscript out of range' at launch


FalseIlyu:
The game trigger an assertion fail at launch, inside `LandBlock::buildVertexList(LandIsland& island)` at line 122 it tries to reach index 266 of the 6th array of the array country,

Behavior appear in commit afb3fc8c117dc485be50e5c2b50dca9cb6bc2c10.

Log here [openblack.log](https://github.com/openblack/openblack/files/8750275/openblack.log)
Here is the stack trace.
```ucrtbased.dll!00007ffb4c46afec() (Source inconnue:0)
ucrtbased.dll!00007ffb4c46af37() (Source inconnue:0)
openblack.exe!std::array<openblack::lnd::LNDMapMaterial,256>::operator[](unsigned __int64 _Pos) Line 559 (c:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\array:559)
openblack.exe!openblack::LandBlock::buildVertexList(openblack::LandIsland & island) Line 122 (d:\openblack\src\3D\LandBlock.cpp:122)
openblack.exe!openblack::LandBlock::BuildMesh(openblack::LandIsland & island) Line 60 (d:\openblack\src\3D\LandBlock.cpp:60)
openblack.exe!openblack::LandIsland::LoadFromFile(const std::filesystem::path & path) Line 90 (d:\openblack\src\3D\LandIsland.cpp:90)
openblack.exe!openblack::Game::LoadLandscape(const std::filesystem::path & path) Line 712 (d:\openblack\src\Game.cpp:712)
openblack.exe!openblack::lhscriptx::FeatureScriptCommands::LoadLandscape(const std::string & path) Line 548 (d:\openblack\src\LHScriptX\FeatureScriptCommands.cpp:548)
openblack.exe!std::invoke<void (__cdecl*&)(std::string const &),std::string const &>(void(*)(const std::string &) & _Obj, const std::string & _Arg1) Line 1505 (c:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\type_traits:1505)
[...]

PR #428, opened 2022-05-22
Static_cast to ensure uint8_t use in array index #427

FalseIlyu:
For some reason even if both altitude and island.GetNoitse are or return uint8_t the sum of both is converted and the value get over 255 creating #427.
This solves it


bwrsandman:
I've got a solution for this bug here:
https://github.com/openblack/openblack/pull/421/files#diff-8203d940c2abae2b98dba22d9841c9d8096e2c53fd6a9d212542a30dd3eb0d3dL119-R131

PR #431, opened 2022-05-24
rendering: Fix invalid data access in entities draw

### LINKED ISSUE #395: Entities not loading or graphic artefact on release build
All information are gathered on a windows build of openblack with vs code through Visual Studio compiler and builder, Configuration is Visual Studio 2022 (vcpkg), screenshot are all from the release build preset.
- Openblack is on ac5c42f.
- vcpkg is on 440075a9.

Visual problems seems to appear only on release and are renderer dependent, with Direct3D12 displaying an empty map.
Openblack_direct12
Other renderers will also display "GPU artifacts" with the most extreme case being redering with OpenGLES where the view get clutered.
Openblack_OpenGLES
Even in Direct3D12 mode artifact can appear, as the bug vary in intensity each launch.

In the log there is a good prevalence of this error message.
`[2022-04-24 13:00:00.642] [game] [error] [L3DMesh.cpp:81] Failed to open L3DSubMesh from file: buffer`
Which if the root is reading from a buffer in an inconsistent state is in line with the way severity will vary from rendering backend and launch.

Here are the log for Release
[openblack.log](https://github.com/openblack/openblack/files/8549501/openblack.log)
Here are the log for debug, where the module appear bug the error are still there.
[openblack.log](https://github.com/openblack/openblack/files/8549508/openblack.log)

This does not appear to be a code issue as my test to track the commit of apparition have led me to version of openblack were i know this problem wasn't there 543c2c987bbab0d7fa0f46baa2300821281b4e46 for example worked fine during my investigation of #382.

Edit : I tried deleting the local repo and cloning it again, the problem persist.

bwrsandman:
`submitDesc` was getting the address of a stack local const variable for the
identity matrix which becomes invalid when the scope ends.

This causes the `u_model[0]` matrix to be invalid at draw time for all
entities and it is seen in release.

Making the identity matrix static moves the data out of the scope and makes it
long-lived solving the problem.

Fixes #395

### DIFF
--- a/src/Renderer.cpp
+++ b/src/Renderer.cpp
@@ -522,7 +522,7 @@ void Renderer::DrawPass(const DrawSceneDesc& desc) const
 				}
 				else
 				{
-					const auto identity = glm::mat4(1.0f);
+					const static auto identity = glm::mat4(1.0f);
 					submitDesc.modelMatrices = &identity;
 					submitDesc.matrixCount = 1;
 				}


FalseIlyu:
Tested the branch, it works entities now appear.

PR #433, opened 2022-05-28
CI: use pull_request_target for format and vcpkg builds to automatic commenting

bwrsandman:
`pull_request_target` differs from `pull_request` in that the workflow is taken from the base.
This prevents security issues of forks stealing secrets but means that PRs for workflows can't be tested
It does, however allow PRs to run tests that use secret tokens


bwrsandman:
Requires checkout to specify
```
        with:
          ref: refs/pull/${{ github.event.pull_request.number }}/merge
```

PR #446, opened 2022-06-04
android: ndk path was broken, ubuntu 22.04 has a new enough ndk by default

bwrsandman:


bwrsandman:
Failing because it's not being tested as a white listed push branch

PR #448, opened 2022-06-04
android: ndk path was broken, ubuntu 22.04 has a new enough ndk by default

bwrsandman:


bwrsandman:
See the results of (Cross Compile CI / build (ubuntu-22.04...) https://github.com/openblack/openblack/actions/runs/2440194981

PR #489, opened 2022-06-10
temple-creature - Draw temple and creature meshes

raffclar:


components/cam/include/CamFile.h
```diff
@@ -0,0 +1,53 @@
...
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include <array>
+#include <string>
+#include <vector>
+
+#include <glm/glm.hpp>
```
bwrsandman, line 16: Don't need to include all of glm for just vec3

components/cam/include/CamFile.h
```diff
@@ -0,0 +1,53 @@
...
+{
+	uint32_t unk1;           // No perceptible difference when altered. Some values cause the game to crash
+	uint32_t movementSpeed;  // The larger this integer, the slower the movement speed
+	uint32_t numberOfPoints; // How many path vertices there are in the file
+};
+
+struct CamPoint
+{
+	glm::vec3 position;
+	glm::vec3 lookAtPosition;
```
bwrsandman, line 31: Since `CameraPath::LoadFromFile` copies this data over to store in a `CameraPoint`, you could use `std::array<float, 3>` which would allow us to have this component exist without an explicit 3rd party dependency like glm.

Instead of the copy, you would use `glm::make_vec3(point.position.data())`

bwrsandman (changes requested):

components/cam/src/CamFile.cpp
```diff
@@ -0,0 +1,84 @@
...
+		float z = 0;
+		CamPoint point = _points[i];
+		output.write(reinterpret_cast<char*>(&point.position.x), sizeof(float));
+		output.write(reinterpret_cast<char*>(&point.position.y), sizeof(float));
+		output.write(reinterpret_cast<char*>(&point.position.z), sizeof(float));
+		output.write(reinterpret_cast<char*>(&x), sizeof(float));
+		output.write(reinterpret_cast<char*>(&y), sizeof(float));
+		output.write(reinterpret_cast<char*>(&z), sizeof(float));
+	}
+	output.close();
```
bwrsandman, line 82: What's going on here? This isn't right...

components/cam/include/CamFile.h
```diff
@@ -0,0 +1,53 @@
...
+#include <vector>
+
+#include <glm/glm.hpp>
+
+namespace openblack::cam
+{
+
+struct CamHeader
+{
+	uint32_t unk1;           // No perceptible difference when altered. Some values cause the game to crash
```
bwrsandman, line 23: It's the file size

bwrsandman:
Closing because this PR has been split into #559, #560, #558,  #557 and #548

The review will continue there.

Please make sure all the changes are accounted for and delete the branch

PR #491, opened 2022-06-18
ecs-systems-locator-490 - Refactor ECS system to use the locator pattern

raffclar:


src/Locator.h
```diff
@@ -9,10 +9,15 @@
 
 #pragma once
 
+#include <ECS/Systems/PathfindingSystem.h>
```
bwrsandman, line 12: Don't use angle brackets for non system includes

src/ECS/Systems/LivingActionSystem.h
```diff
@@ -13,24 +13,38 @@
...
-	bool VillagerCallExitState(components::LivingAction& action, components::LivingAction::Index index) const;
-	int VillagerCallOutOfAnimation(components::LivingAction& action, components::LivingAction::Index index) const;
-	bool VillagerCallValidate(components::LivingAction& action, components::LivingAction::Index index) const;
-
-private:
-	LivingActionSystem();
+	                            VillagerStates dst) const override;
+	bool VillagerCallExitState(components::LivingAction& action, components::LivingAction::Index index) const override;
+	int VillagerCallOutOfAnimation(components::LivingAction& action, components::LivingAction::Index index) const override;
+	bool VillagerCallValidate(components::LivingAction& action, components::LivingAction::Index index) const override;
```
bwrsandman, line 48: Private functions do not need to be virtual

src/ECS/Systems/CameraBookmarkSystem.h
```diff
@@ -18,21 +18,28 @@
...
+	void Update(const std::chrono::microseconds& dt) const override;
 
-	void SetBookmark(uint8_t index, const glm::vec3& position) const;
-	void ClearBookmark(uint8_t index) const;
+	[[nodiscard]] const std::array<entt::entity, 8>& GetBookmarks() const override { return _bookmarks; }
 
-private:
-	CameraBookmarkSystem();
+	void SetBookmark(uint8_t index, const glm::vec3& position) const override;
+	void ClearBookmark(uint8_t index) const override;
```
bwrsandman, line 40: Private functions don't need to be virtual

bwrsandman (changes requested):
Not sure how but your changes broke Enums that are named None. Probably because a None got a value somewhere.
This can probably be fixed by reordering includes or renaming None to `NONE` (Not a fan of renaming)

src/ECS/Systems/CameraBookmarkSystem.h
raffclar: These are public before the refactor.

src/Locator.h
raffclar: Nice catch

src/ECS/Systems/LivingActionSystem.h
raffclar: Same here. These are all public before the refactor

src/Resources/MeshId.h
```diff
@@ -9,6 +9,8 @@
 
 #pragma once
 
+#include <entt/entt.hpp>
```
bwrsandman, line 12: entt/core/hashed_string.hpp

src/Locator.h
```diff
@@ -11,17 +11,35 @@
...
+
+namespace ecs::systems
+{
+class RenderingSystemInterface;
+class DynamicsSystemInterface;
+class CameraBookmarkSystemInterface;
+class LivingActionSystemInterface;
+class TownSystemInterface;
+class PathfindingSystemInterface;
+} // namespace ecs::systems
```
bwrsandman, line 32: Awesome!

bwrsandman (changes requested):
Is it a good idea to have the Abstract base classes in the same headers as the default provided definitions?

src/Resources/MeshId.h
raffclar: > Is it a good idea to have the Abstract base classes in the same headers as the default provided definitions?

No it isn't. We don't have multiple implementations and we have forward declarations now so there's no performance to be gained.

Might take a bit of work to do but probably best to do it now before we have 100 systems.

raffclar: > entt/core/hashed_string.hpp

Thanks

bwrsandman (approved):

src/Resources/MeshId.h
raffclar: > > Is it a good idea to have the Abstract base classes in the same headers as the default provided definitions?
> 
> No it isn't. We don't have multiple implementations and we have forward declarations now so there's no performance to be gained.
> 
> Might take a bit of work to do but probably best to do it now before we have 100 systems.

@bwrsandman #492

PR #499, opened 2022-06-26
Refactor File-related classes to a separate component

raffclar:


src/3D/L3DAnim.cpp
```diff
@@ -13,7 +13,7 @@
 #include <glm/gtx/matrix_interpolation.hpp>
 #include <spdlog/spdlog.h>
 
-#include "Common/FileSystem.h"
```
raffclar, line 16: @bwrsandman 
`#include "FileSystem.h"` feels a bit too open. Should we go for `#include "file_utils/FileSystem.h"`?

bwrsandman (changes requested):
This is not done right.

Components in the `component/` directory refer to pieces that can be used separately from openblack in regards to Black & White modding/reimplmentation.

They are here to be independent from the main project to have so they can be used to make mods, plugins or other implementations.

As you can see the current components are all for a different file format. They are paired with minimal example uses of them in the `apps` directory as well as being used in openblack.

Someone coming here wanting to make a new type of villager can use `pack`, `l3d` and `anm` to make a blender plugin as they're separate with minimal dependencies.

file_utils doesn't fit in the components directory. These are util classes of openblack. They don't add anything new to anyone wanting to mod Black & White. They just make our file easier inside of openblack

raffclar:
> This is not done right.
> 
> Components in the `component/` directory refer to pieces that can be used separately from openblack in regards to Black & White modding/reimplmentation.
> 
> They are here to be independent from the main project to have so they can be used to make mods, plugins or other implementations.
> 
> As you can see the current components are all for a different file format. They are paired with minimal example uses of them in the `apps` directory as well as being used in openblack.
> 
> Someone coming here wanting to make a new type of villager can use `pack`, `l3d` and `anm` to make a blender plugin as they're separate with minimal dependencies.
> 
> file_utils doesn't fit in the components directory. These are util classes of openblack. They don't add anything new to anyone wanting to mod Black & White. They just make our file easier inside of openblack

If we to leverage some of our common classes in a component (in my case, audio) how could we refactor them so that they sit outside of the src directory?
The relationship between openblack_lib and components is one way

bwrsandman:
Is there a way to have the audio component have as little dependencies as possible?
What does it need at the very least?

PR #505, opened 2022-07-01
l3d: Fix flags of submesh

### LINKED ISSUE #185: LOD system not implemented
This is actually prefferable, so perhaps a toggle?

Here are gifs, as it is difficult to explain. 
runblack: https://imgur.com/KWyWjzM
openblack: https://imgur.com/dmh4Ehl

The terrain does not deform to a lower poly, the trees don't pop to their 2d cutout, and buildings do not fade with distance. Can't test with entities yet, because they don't exist.

bwrsandman:
The LOD field was wrong. Firstly, it is at the other end of the dword, second, it is a bitmask for lods 1, 2, 4.
The Status field was offset by one bit and it encodes information about how the object morphs
in the case of building and a graveyard being filled.
Disabled drawing objects that have a status because no status is implemented yet.
Also fixed the meshviewer not showing physics meshes.

TODO:
- [x] Select an LOD for the trees (just go with LOD 1). Closes  #185 

image

### DIFF
--- a/apps/l3dtool/l3dtool.cpp
+++ b/apps/l3dtool/l3dtool.cpp
@@ -229,7 +229,7 @@ int PrintMeshHeaders(openblack::l3d::L3DFile& l3d)
 		{
 			result += "hasBones";
 		}
-		result += "|LOD" + std::to_string(flags.lod);
+		result += "|lodMask=" + std::to_string(flags.lodMask);
 		result += "|status=" + std::to_string(flags.status);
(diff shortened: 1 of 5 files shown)


bwrsandman:
The changes to LOD can allow us to select only one LOD for the trees. This is not done in this PR though.

handsomematt (approved):
These flags have alluded me for years and years, excellent find!

PR #509, opened 2022-07-04
l3d: Extract footprint mesh and texture

bwrsandman:
image
image
image

TODO: 
- [x] Sometimes there's more than one footprint. There are two meshes that have 4 in the mesh pack.
- [x] Vanilla draws footprints onto a framebuffer that the landscape samples from instead of the texture array. We might be able to get away with something more like a decal. There is z-fighting and drawing the footprint shouldn't write to depth.
- [x] Some footprints might not be shadows but lights in some cases
- [x] Switching lands breaks the rendering
- [x] Works in OpenGL but not Vulkan (Failed depth test?)
- [x] Move Pass code out of DrawScene
- [x] SegFault on exit in Vulkan
- [x] Add a TODO saying that the footprint framebuffer doesn't need to be updated each frame
- [x] The shadowy part of the village centre is not the same in vanilla

### DIFF
--- a/components/l3d/src/L3DFile.cpp
+++ b/components/l3d/src/L3DFile.cpp
@@ -22,7 +22,8 @@
  *         extra data offset - offset of the point block (see below)
  *         footprint data offset - offset of the footprint block (see below)
  *
- * There are a few optional blocks. Their presence can be check for using the flags field
+ * There are a few optional blocks. Their presence can be check for using the
+ * flags field
  * The footprint block is present when ContainsLandscapeFeature is set
  * The UV2 block is present when ContainsUV2 is set
  * The name block is present when ContainsNameData is set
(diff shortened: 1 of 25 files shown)


bwrsandman:
The footprint might modify the height map?

handsomematt:
Yeah I believe so, or at least it acts like a modifier whilst it's placed down.

components/l3d/src/L3DFile.cpp
```diff
@@ -47,7 +48,25 @@
...
+ *         zero - 4 bytes
+ * - number of entries * variable size entries containing:
+ *         possible packed color - 4 bytes
+ *         zero - 4 bytes
+ *         triangle count - 4 bytes
+ *         triangle * count * 3 pairs of floats reprensenting world space and
+ * texture space
+ *         texture width * height of 2 byte grayscale values in [0, 0xFFFF]
+ *         zero - 12 bytes
+ * - 12 byte header containing:
```
bwrsandman, line 66: footer

handsomematt (approved):
Perfect. Looks brilliant and the implementation is fantastic.

PR #513, opened 2022-07-26
Morph with Terrain

bwrsandman:
This PR adds the functionality to morph objects with the terrain height map.
The objects that do this are documented in https://github.com/openblack/openblack/wiki/3D-Object-Types

image
image

Code clean-up and refactor:
* Fixed the Graveyard info.dat enum value
* Added filter type for texture2d in case is it better to point sample height map
* Cleaned-up terrain block mesh generation to be more readable and to expand upon if we decide to merge the meshes into one large one
* The instancing code in Rendering System now uses an unordered map for faster look-ups

I chose not to sample the height map in a shader as this would be a waste of cycles since the height map is static.
Objects are sampled each frame because the meshes are shared and instance drawn.

The attribute for morphing with terrain was added to the instancing code in RenderingSystem because there is no ordering.
I'm not 100% happy with this and would have preferred to have the instances sorted and storing the range of morphable objects.

TODO:
- [x] Depends on #517 and needs to add `MorphWithTerrain` to Food type Pots
- [x] Look-up which tribes morph their wonders
- [x] Add `MorphWithTerrain` component in `MobileStaticArchetype::Create(..., MobileStaticInfo::PhysicalShield, ...)`

### DIFF
--- a/src/ECS/Components/MorphWithTerrain.h
+++ b/src/ECS/Components/MorphWithTerrain.h
@@ -1,0 +1,37 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2022 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+namespace openblack::ecs::components
+{
+
+/// Meshes with this component will have their vertices match terrain height maps
+///
+/// They correspond with vanilla Get3DType of MORPHABLE or CITADEL
+/// The following Objects should be created with this component:
+/// * BigForest
+/// * Graveyard
+/// * PileFood (not wood)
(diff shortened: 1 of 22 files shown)


src/ECS/Components/MorphWithTerrain.h
```diff
@@ -0,0 +1,37 @@
...
+{
+
+/// Meshes with this component will have their vertices match terrain height maps
+///
+/// They correspond with vanilla Get3DType of MORPHABLE or CITADEL
+/// The following Objects should be created with this component:
+/// * BigForest
+/// * Graveyard
+/// * TODO: PileFood (not wood)
+/// * StoragePit x2 meshes
```
bwrsandman, line 22: Probably has to do with the PileFood

bwrsandman:
Missing bbe9410c3a2e488bb084b35a79565cf0227d11cb

raffclar:
This is really good. Ping me asap when you need a review

raffclar (approved):
Really nice. I like the way you did the shader macroing

PR #517, opened 2022-07-27
Piles and storage pit

bwrsandman:
This PR does three things:
* Decodes the Extra Metrics L3D block.
* Implements the `CREATE_POT` script command
* Adds the Wood and Food piles to storage pits

image
image
image

### DIFF
--- a/src/ECS/Components/StoragePit.h
+++ b/src/ECS/Components/StoragePit.h
@@ -1,0 +1,25 @@
+/*****************************************************************************
+ * Copyright (c) 2018-2022 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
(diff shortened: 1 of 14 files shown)


bwrsandman:
The fail on x86 is unrelated and happens on every test. I suspect a bad cache.

src/ECS/Components/StoragePit.h
```diff
@@ -0,0 +1,25 @@
...
+#include <array>
+
+#include <entt/fwd.hpp>
+
+namespace openblack::ecs::components
+{
+
+struct StoragePit
+{
+	std::array<entt::entity, 5> woodPiles;
```
raffclar, line 21: Where did the 5 come from?

raffclar (commented):
Really good. Just 1 q about the constant in one of the components

raffclar (approved):

src/ECS/Components/StoragePit.h
bwrsandman: There are 5 separate piles here
image

raffclar: Ah ok got it

PR #449, opened 2022-06-05
Screenshot

bwrsandman:
Re-opening #418 but with a openblack branch so that action changes to `pull_request_target` can be tested

Requires two secrets from @handsomematt :
- [x] Game Data credentials
- [x] imgbb API key
- [x] Depends on #528 for fix to engine shutdown error code
- [x] Depends on #523 for loading specific scenes

Proof of concept: https://github.com/openblack/openblack/pull/486#issuecomment-1203315411

In test PR bad2b789f7d0b1336078e8500c752caaec4f6927 worked on top of 6da1b4a7

PR #528, opened 2022-08-02
Fix d3d11 release shutdown bug

bwrsandman:
There is a bug in d3d11 when shutting down bgfx. A failure happens in the driver code and it happens only when `instanceUniformBuffer` is used.

This bug prevents #449 from completing as the game exits non-cleanly.


bwrsandman:
Seems like there are still crashes in the screenshot actions

bwrsandman:
appears to be non-necessary with #529

Issue #174, opened 2019-12-10
Add Windows ARM as a supported target platform


handsomematt:
Would be a nice feature, Linux ARM could be done too but is less of a widely used target.

NeroBurner:
I'd say linux arm is a bit more widespread than you think
- https://www.pine64.org/pinebook-pro/
- https://www.raspberrypi.org/
- https://www.nvidia.com/en-us/shield/

bwrsandman:
> android

bwrsandman:
All our VCPKG dependencies now support windows ARM, linux ARM and android ARM as part of my push for #262

PR #520, opened 2022-07-29
Enable x86, arm64 and webasm cross compile tests

bwrsandman:
This reverts commit 4b0df2ba8831bc1b27dad9096ea74e06a81f792a.

Closes #519, #174 and #533

Issue #519, opened 2022-07-29
Re-enable x86 tests


bwrsandman:
Some time last week the x86 tests broke not through any changes of code on our end. It seems like a github workflow environment problem.

>  [1/133] Building CXX object externals\CMakeFiles\imgui_user.dir\Debug\imgui_user\imgui_widget_flamegraph.cpp.obj
  FAILED: externals/CMakeFiles/imgui_user.dir/Debug/imgui_user/imgui_widget_flamegraph.cpp.obj 
  C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1432~1.313\bin\Hostx64\x86\cl.exe  /nologo /TP -DCMAKE_INTDIR=\"Debug\" -ID:\a\openblack\openblack\externals\imgui_user -external:ID:\a\openblack\openblack\cmake-build-presets\ninja-multi-vcpkg\vcpkg_installed\x86-windows\include -external:W0 /DWIN32 /D_WINDOWS /GR /EHsc /Zi /Ob0 /Od /RTC1 -MDd -std:c++20 /showIncludes /Foexternals\CMakeFiles\imgui_user.dir\Debug\imgui_user\imgui_widget_flamegraph.cpp.obj /Fdexternals\CMakeFiles\imgui_user.dir\Debug\imgui_user.pdb /FS -c D:\a\openblack\openblack\externals\imgui_user\imgui_widget_flamegraph.cpp
  Error: D:\a\openblack\openblack\externals\imgui_user\imgui_widget_flamegraph.h(31): fatal error C1083: Cannot open include file: 'climits': No such file or directory
  [2/133] Building CXX object externals\CMakeFiles\imgui_user.dir\Debug\imgui_user\imgui_user.cpp.obj
  FAILED: externals/CMakeFiles/imgui_user.dir/Debug/imgui_user/imgui_user.cpp.obj 
[...]

PR #550, opened 2022-08-25
ci: Add arm cross compile tests

bwrsandman:
implements #174


bwrsandman:
arm-windows suffering from same issue as x86-windows #519 `Cannot open include file: 'float.h': No such file or directory`

bwrsandman:
#520

PR #572, opened 2022-12-03
vcpkg: Use bgfx from repo

bwrsandman:


handsomematt:
lgtm so long as its targeting Microsoft/vcpkg

PR #586, opened 2023-05-18
Action map for input

bwrsandman:
This PR is the first in two or more parts for re-doing the camera controller model.

This bit brings the vanilla input mapping as a service.
The part for rebinding keys is not yet done. This would require a menu, defined Key names and icons and the ability to save and fetch them from the system like vanilla (#293 is close but is using a totally new input mapping strategy).

Vanilla has a total of 64 inputs that can be bound but it also has inputs that can't be rebound such as double clicking. There is also mouse position and delta.

As a proof of concept, I re-worked the current camera to use the input map and no longer use SDL2. The result works as before but the code is far from elegant since it was written with the event callbacks in mind. Since moving the camera up and down and using the ctrl key is not in the original keymapping, I re-assigned moving up and down to `SPACE` and `SHIFT + SPACE`

The Camera model will be completely re-written in a future PR.

I'm using contracts from C++20 and variants from C++17.

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -355,6 +355,7 @@ bool Game::Update()
 		{
 			_eventManager->Create<SDL_Event>(e);
 		}
+		_camera->HandleActions();
 	}
 
 	if (!this->_config.running)
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -346,7 +346,10 @@ bool Game::Update()
 	// Input events
 	{
 		auto sdlInput = _profiler->BeginScoped(Profiler::Stage::SdlInput);
-		Locator::gameActionSystem::value().Frame();
+		if (!this->_gui->StealsFocus())
+		{
+			Locator::gameActionSystem::value().Frame();
+		}
 		SDL_Event e;
 		while (SDL_PollEvent(&e) != 0)
 		{
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -131,7 +131,6 @@ Game::Game(Arguments&& args)
 		// If gui captures this input, do not propagate
 		if (!this->_gui->ProcessEvents(event))
 		{
-			this->_camera->ProcessSDLEvent(event);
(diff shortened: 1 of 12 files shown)


src/Input/GameActionMapInterface.h
```diff
@@ -0,0 +1,193 @@
...
+	template <typename... Args>
+	    requires(... && (std::is_same_v<Args, BindableActionMap> || std::is_same_v<Args, UnbindableActionMap>))
+	bool GetChangedAll(Args... actions) const
+	{
+		return (GetChanged(actions) && ...);
+	}
+
+	template <typename... Args>
+	    requires(... && (std::is_same_v<Args, BindableActionMap> || std::is_same_v<Args, UnbindableActionMap>))
+	bool GetRepeatAll(Args... actions) const
```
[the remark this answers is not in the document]
bwrsandman: Bad suggestion and not clion is giving a different format

src/Game.cpp
```diff
@@ -51,6 +51,8 @@
 #include "GameWindow.h"
 #include "Graphics/FrameBuffer.h"
 #include "Graphics/Texture2D.h"
+#include "Input/GameActionMap.h"
```
bwrsandman, line 54: Not a big fan of including this. Maybe a factory function in GameActionMapInterface would be better.

src/Game.cpp
```diff
@@ -478,6 +486,7 @@ bool Game::Initialize()
 {
 	Locator::resources::emplace<resources::Resources>();
 	Locator::rng::emplace<RandomNumberManagerProduction>();
+	Locator::gameActionSystem::emplace<input::GameActionMap>();
 	ecs::systems::InitializeGame();
```
bwrsandman, line 490: Should happen in locator.cpp

raffclar (approved):

FalseIlyu (approved):

PR #589, opened 2023-06-02
Use pure config of run-vcpkg@11

bwrsandman:
This fixes windows builder defaulting to vcpkg installed with msvc. It also unifies the caches and seems to run much faster.

Ref https://github.com/lukka/run-cmake/issues/113


handsomematt (approved):
awesome

Issue #400, opened 2022-04-26
Parse level files within the resource loader


raffclar:
Once #287 is merged we need to parse the level files within the Level resource loader so that we can display additional info from `START_GAME_MESSAGE` and `ADD_GAME_MESSAGE_LINE` statements. 

If the level cannot be parsed it should be greyed out.
if the level is parsed and a START_GAME_MESSAGE statement is present then the first argument should be used in the menu select.

bwrsandman:
The implementation should be able to work even if there are no playground maps and no Land*.txt quest maps.
This means the PR should remove Land 2-5 and Playgrounds from mock. Land 1 should stay

PR #593, opened 2023-06-13
Level loader parsing

FalseIlyu:
Proposal for issue #400.


bwrsandman:
Why are you removing mock lands?

FalseIlyu:
At this moment, the seperation between campaign and playground mission is handled by looking if the land number is > 0.
Giving result like below.
image
Is this behavior acceptable or should the campaign/not campaign nature be determined by the script place in the game filesystem ?

bwrsandman:
Wow! Great progress!

It would be worth it to investigate what vanilla black & white does to avoid putting those into the skirmish maps.
What are your thoughts about this?

bwrsandman:
Make sure you run the unit tests. They seem to be failing.
```bash
ctest --preset ninja-multi-vcpkg-debug
```

FalseIlyu:
> Wow! Great progress!
> 
> It would be worth it to investigate what vanilla black & white does to avoid putting those into the skirmish maps. What are your thoughts about this?

Vanilla seems to just load any map script inside the Playgrounds into the skirmish menu (campaign maps will result in an instant defeat). I think this system of looking the land number was a mistake and put an unecessary constraint on map makers while just using folder placement give an incentive for user to not just put all their skirmish map in the main script folder in a big mess.

> Make sure you run the unit tests. They seem to be failing.
> 
> ```shell
> ctest --preset ninja-multi-vcpkg-debug
> ```

It seems that removing the mocks create problems with the test (stop at loading the map), i just dropped the commit that removed them.

src/Resources/Loaders.h
```diff
@@ -59,7 +59,8 @@ struct L3DAnimLoader final: BaseLoader<L3DAnim>
 
 struct LevelLoader final: BaseLoader<Level>
 {
-	[[nodiscard]] result_type operator()(FromDiskTag, const std::string& name, const std::filesystem::path& path) const;
+	[[nodiscard]] result_type operator()(FromDiskTag, const std::string& name, const std::filesystem::path& path,
+	                                     bool isCampaign) const;
```
bwrsandman: For better readability, strong typing and general style, I would always replace a `bool` with an `enum class` in cases like these.

src/Debug/Gui.cpp
```diff
@@ -452,27 +452,28 @@ bool Gui::ShowMenu(Game& game)
...
-					ImGui::SetTooltip("%s", path.generic_string().c_str());
+					ImGui::SetTooltip("%s", description.c_str());
 				}
 			};
 
 			ImGui::MenuItem("Story Islands", nullptr, false, false);
 			for (auto& level : campaigns)
 			{
-				menuItem(level->GetName(), level->GetScriptPath());
+				menuItem(level->GetName(), level->GetScriptPath(), level->GetScriptPath().generic_string(), level->IsValid());
```
bwrsandman, line 470: Is this on purpose to have the path instead of the description?
Is the `IsValid` call making the item disabled?

bwrsandman (changes requested):
Great work so far. It's simple and seems pretty efficient.

I haven't tested it yet but it looks like my only issues are about naming.

bwrsandman:
You'll need to update your branch with master to fix the windows error. I recommend rebasing if you're comfortable doing that.

src/Debug/Gui.cpp
FalseIlyu: > Is this on purpose to have the path instead of the description?
It was on purpose but it was misktaken, i used the path as the description are extracted from ADD_GAME_MESSAGE_LINE inside script file and campaign map don't have this instruction. Now the loader build the description and you're just supposed to get it when needed.

> Is the IsValid call making the item disabled?
Yes, if i can't find LOAD_LANDSCAPE instruction the item is disabled.

src/Level.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+	return _description;
+}
+
+bool Level::IsValid() const
+{
+	return _isValid;
+}
+
+void Level::ParseLevel(const std::filesystem::path& path, bool& isValid, std::string& levelName, std::string& description)
```
bwrsandman, line 48: Do you think it would make more sense to return the validity of the file?

This function also gets 3/4 things needed to create a Level object, does it make sense to turn this into a factory function given the name.

src/Level.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <fstream>
+
+using namespace openblack;
+
+std::string extractQuote(std::string& string)
+{
+	size_t const first(string.find('\"'));
+	size_t const second(string.find('\"', first + 1));
+	return string.substr(first + 1, second - first - 1);
+}
```
bwrsandman, line 21: I think @raffclar wrote a string utils, that should go there.

src/Level.h
```diff
@@ -11,26 +11,41 @@
...
+
+	Level(std::string name, std::filesystem::path path, std::string description, LandType landType, bool isValid = false)
 	    : _name(std::move(name))
 	    , _scriptPath(std::move(path))
-	    , _isCampaign(isCampaign)
-	{
-	}
+	    , _description(std::move(description))
+	    , _landType(landType)
+	    , _isValid(isValid) {};
```
bwrsandman, line 33: The definition can also move to the cpp file.
This way we can remove the `<utility>` include from a public header.

src/Resources/Loaders.h
```diff
@@ -60,7 +60,7 @@ struct L3DAnimLoader final: BaseLoader<L3DAnim>
 struct LevelLoader final: BaseLoader<Level>
 {
 	[[nodiscard]] result_type operator()(FromDiskTag, const std::string& name, const std::filesystem::path& path,
-	                                     bool isCampaign) const;
+	                                     Level::LandType isCampaign) const;
```
bwrsandman, line 63:
```suggestion
	                                     Level::LandType landType) const;
```

bwrsandman (changes requested):
Few more changes.
I ran your branch and I have some questions:
1. Does it make sense to have the path as tooltips on story but not on playground? What's the purpose of showing the path?
2. Invalid maps look indistinguishable from header titles (see `map` and `Story Island`
3. Did you consider sorting the entries?
4. If I try loading dance, there is an entry for load lanscape but the file does not exist, shouldn't this be invalid?

image

bwrsandman:
When you are done working and just doing reviews, you can set the PR as ready for review.

src/Enums.h
```diff
@@ -3406,4 +3406,10 @@ enum class CreatureType : int32_t
 	_COUNT = 18
 };
 
+enum class LandType : uint32_t
+{
+	Skirmish = 0,
+	Campaign = 1,
+};
+
```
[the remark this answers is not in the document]
bwrsandman: Also, the underlying type doesn't have to be 32 bits and there's no need to set values explicitly.

src/Level.h
```diff
@@ -14,23 +14,32 @@
...
-	[[nodiscard]] bool IsCampaign() const { return _isCampaign; };
+	static Level ParseLevel(const std::filesystem::path& path, Level::LandType landType);
 
 private:
 	std::string _name;
 	std::filesystem::path _scriptPath;
-	bool _isCampaign;
+	std::string _description;
+	LandType _landType;
+	/// @brief Does the loaded level script contains at least a LOAD_LANDSCAPE instruction ?
```
bwrsandman, line 42:
```suggestion
	/// @brief The loaded level script contains at least a LOAD_LANDSCAPE instruction
```

bwrsandman (approved):

bwrsandman:
You should mark the PR as ready to review to let me know if I can merge

src/Common/StringUtils.cpp
```diff
@@ -69,3 +69,11 @@ std::string openblack::string_utils::Capitalise(const std::string& string)
 
 	return parts;
 }
+
+/// Extract a substring of the characters in between the first two quote of a string
```
bwrsandman, line 73: This should be in the header

bwrsandman:
@FalseIlyu I might have to revert this merge because the change seems to not work on linux and possibly other platforms:

1. The sorts cause the game to hang when opening the Load Island menu.
2. All lands are greyed out.

Can you let me know if you're having a look?

src/Level.cpp
```diff
@@ -0,0 +1,110 @@
...
+	std::fstream levelFile;
+	levelFile.open(path, std::ios::in);
+	while (!levelFile.eof())
+	{
+		std::string line;
+		std::getline(levelFile, line);
+		if (!isValid && line.find(loadLandscapeLine) != std::string::npos)
+		{
+			std::filesystem::path const landscapePath(string_utils::ExtractQuote(line));
+			isValid = Game::Instance()->GetFileSystem().Exists(landscapePath);
```
bwrsandman, line 92: landscapePath here is in Windows format and fails on non-windows platforms.
You can fix the path using `FileSystem::FixPath` before pluging the result into Exists

PR #597, opened 2023-06-13
CI: Update clang-tidy-review and use split workflow config

bwrsandman:


bwrsandman:
Can't get the split files to run

PR #608, opened 2023-06-14
CI: Remove redundant specifications in run-cmake

bwrsandman:


bwrsandman:
Windows build is failing due to environment variables not being set in build steps: https://github.com/lukka/run-cmake/issues/115

PR #612, opened 2023-06-19
Fix landscape path before use.

FalseIlyu:
Should fix the bug found by @bwrsandman in #593 .

### DIFF
--- a/src/Debug/Gui.cpp
+++ b/src/Debug/Gui.cpp
@@ -452,12 +452,10 @@ bool Gui::ShowMenu(Game& game)
 				}
 			});
 
-			std::sort(campaigns.begin(), campaigns.end(), [](const Level* level1, const Level* level2) {
-				return &level1 < &level2;
-			});
-			std::sort(playgrounds.begin(), playgrounds.end(), [](const Level* level1, const Level* level2) {
-				return &level1 < &level2;
-			});
+			std::sort(campaigns.begin(), campaigns.end(),
+			          [](const Level* level1, const Level* level2) { return &level1 < &level2; });
+			std::sort(playgrounds.begin(), playgrounds.end(),
+			          [](const Level* level1, const Level* level2) { return &level1 < &level2; });
 
 			auto menuItem = [&game](const auto& label, const std::filesystem::path& path, const std::string description,
 			                        bool validLevel) {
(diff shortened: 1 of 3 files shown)


bwrsandman:
Still crashing due to the two sort calls

bwrsandman (changes requested):
The sorting function no longer crashes but it still bugged.

Additionally, I think you should be using submenus instead of MenuItem + Separator:

```
			if (ImGui::BeginMenu("Playground Islands"))
			{
				for (auto& level : playgrounds)
				{
					menuItem(level->GetName(), level->GetScriptPath(), level->GetDescription(), level->IsValid());
				}
				ImGui::EndMenu();
			}
```

Another issue is that the descriptions of Campaign levels is empty and it results in useless empty tooltips. Is it possible to hide or replace with something when empty?

src/Debug/Gui.cpp
```diff
@@ -452,8 +452,10 @@ bool Gui::ShowMenu(Game& game)
 				}
 			});
 
-			std::sort(campaigns.begin(), campaigns.end());
-			std::sort(playgrounds.begin(), playgrounds.end());
+			std::sort(campaigns.begin(), campaigns.end(),
+			          [](const Level* level1, const Level* level2) { return &level1 < &level2; });
+			std::sort(playgrounds.begin(), playgrounds.end(),
+			          [](const Level* level1, const Level* level2) { return &level1 < &level2; });
```
bwrsandman, line 458: This is worse. You're sorting addresses of addresses.

src/Level.cpp
```diff
@@ -106,5 +106,5 @@ Level Level::ParseLevel(const std::filesystem::path& path, Level::LandType landT
 
 bool operator<(const Level& first, const Level& second)
 {
-	return first.GetName() < second.GetName();
+	return (first.GetName() < second.GetName());
 }
```
bwrsandman, line 110: Get rid of this. It doesn't make sense other than sorting and we only sort in the UI.

src/Debug/Gui.cpp
```diff
@@ -463,24 +465,30 @@ bool Gui::ShowMenu(Game& game)
 				}
 				if (ImGui::IsItemHovered())
 				{
-					ImGui::SetTooltip("%s", description.c_str());
+					if (!description.empty())
```
bwrsandman, line 468: You can merge these two conditions

bwrsandman (changes requested):

bwrsandman (approved):

Issue #470, opened 2022-06-07
LandBlock: Vertex blending should be moved to shader in the renderer


bwrsandman:
src/3D/LandBlock.cpp#L133

Issue #218, opened 2020-02-12
Performance loss at 'Draw Island'


SulpherStaer:
In 'debug', any 'small hitch' is greatly compounded versus 'release'.

The example here is the FPS gain / loss with Draw Island enabled or disabled:

|Debug | Release|
| --- | --- |
| ![image](https://user-images.githubusercontent.com/5037244/74365640-c9a72e80-4dce-11ea-9b25-15a2edc33517.png)| ![image](https://user-images.githubusercontent.com/5037244/74365714-ec394780-4dce-11ea-8b7f-3c7c4c19fd7d.png) |
| ![image](https://user-images.githubusercontent.com/5037244/74365666-d166d300-4dce-11ea-8434-dbefd7f2d6c7.png) | ![image](https://user-images.githubusercontent.com/5037244/74365740-f5c2af80-4dce-11ea-81a8-41c162ae51ab.png) |

bwrsandman:
Draw island does a lot of draw calls that aren't instanced. Instancing them would likely remove this perf loss on the CPU.
That being said, instanced drawing would make it impossible to do culling, which I don't think we even do.

bwrsandman:
duplicate of #81

bwrsandman:
Testing today on 996ac74e2a5165363e8fc46cd8c24c98f6bfa060, it seems like this issue is resolved.
Encoding draws for island on debug takes about 20% the time as encoding drawing entities.
![image](https://github.com/openblack/openblack/assets/1013356/cf51d290-8fb7-4aab-ad65-5b64f11947e8)



**Environment**
```
 OS: Arch Linux 
 Kernel: x86_64 Linux 6.8.7-arch1-1
 DE: GNOME 46.0.1
 CPU: AMD Ryzen 9 5950X 16-Core @ 32x 3.4GHz
 GPU: NVIDIA GeForce RTX 2080
 RAM: 9125MiB / 64209MiB
```

PR #615, opened 2023-06-28
Draw the terrain using an instanced call

### LINKED ISSUE #81: Render terrain with instanced draw
Currently, terrain is [drawn in a for loop](https://github.com/openblack/openblack/blob/71eb61dca92eff61924e1687bb12248d3667533d/src/3D/LandIsland.cpp#L203).
Each block has a uniform uploaded and is drawn before the next one.

Uploading uniforms between draw calls can stall the GPU and is not recommended.

Instead, it would be better to upload all uniforms at initialization time in a uniform array.
The draw calls should be replaced with a single instanced draw call and the uniform array should be looked up with instance id.

FalseIlyu:
If completed should close #81 #470 and #218, replace #613.

As of now the uniforms in the shader have been set up as instance input (islandExtent and u_skyAndBump) probably should go back to uniform.
In the shader the height of the vertex is computed by looking at the heighmap (given through a 2DSampler) and multiplying the height map value by the heightunit (given though an instance input, it probably could be turned into a uniform).
At the moment a single call is made with the vertexBuffer of an arbitray block as the blending data is given through data inside the LandVertex it means all block share the same texture.
WaterAlpha is computed arbitraly inside the shader by clamping altitude between 0 and 1.

image

- [ ] Make texture blending works in an instanced call.
- [ ] Make waterAlpha use cell properties in shader

### DIFF
--- a/assets/shaders/fs_terrain.sc
+++ b/assets/shaders/fs_terrain.sc
@@ -1,22 +1,22 @@
-$input v_texcoord0, v_texcoord1, v_weight, v_materialID0, v_materialID1, v_materialBlend, v_lightLevel, v_waterAlpha, v_distToCamera
+$input v_texcoord0, v_texcoord1, v_weight, v_materialID0, v_materialID1, v_materialBlend, v_lightLevel, v_waterAlpha, v_distToCamera, v_skyABump
 
 #include <bgfx_shader.sh>
 
 #define M_PI 3.1415926535897932384626433832795
 
-SAMPLER2DARRAY(s0_materials, 0);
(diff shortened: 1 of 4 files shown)


bwrsandman:
Closing as we discussed: It's a little compilicated to do the height map and the triangle rotations necessary

FalseIlyu:
> Closing as we discussed: It's a little compilicated to do the height map and the triangle rotations necessary

Maybe you should unmark #81 as a good first issue (maybe as a TODO ? with help wanted ?)

PR #616, opened 2023-06-30
Decoupling the LandIsland from Game to the Locator respecting the Locator Pattern.

### LINKED ISSUE #591: Decouple LandIsland from Game interface
Part of a push to remove the Game singleton.

Use the locator pattern to get access to the LandIsland handle.

Example usage
```c++
// Current
#include "Game.h"
#include "LandIsland.h"
altitude = Game::Instance()->GetLandIsland().GetHeightAt(glm::xz(position3d));

// Better
#include "Locator.h"
#include "LandIsland.h"
altitude = Locator::terrainSystem::value().GetHeightAt(glm::xz(position3d));
```

Sofistico:
Closes #591
Was changed so that the Game instance is less of a singleton holder of instances, done by moving all references to Game::Instance()->GetLandInsland() to the Locator class.
The initialization of the land has also been moved to the LoadLandInsland()


src/Locator.h
```diff
@@ -44,5 +45,6 @@ struct Locator
 	using livingActionSystem = entt::locator<ecs::systems::LivingActionSystemInterface>;
 	using townSystem = entt::locator<ecs::systems::TownSystemInterface>;
 	using pathfindingSystem = entt::locator<ecs::systems::PathfindingSystemInterface>;
+	using terrainSystem = entt::locator<LandIsland>;
```
bwrsandman, line 48: This is not an interface like the other systems. Please have a look at the others to see how they're done.

We're going to want an abstract class interface that Landisland implements and which allows us to easily swap out for another interface in the future.

bwrsandman (changes requested):
Great work so far. Have a look at how we do interfaces. 
There should also be more clean-up of headers and the coupling of game thanks to this change so have a look where game is included or used and doesn't need to anymore.

I also encourage you to look for other bits of game which can be decoupled.

src/Game.cpp
```diff
@@ -815,9 +815,9 @@ void Game::LoadMap(const std::filesystem::path& path)
 void Game::LoadLandscape(const std::filesystem::path& path)
 {
 	ecs::systems::InitializeLevel();
-	if (_landIsland)
+	if (Locator::terrainSystem::has_value())
 	{
-		_landIsland.reset();
+		Locator::terrainSystem::reset();
```
bwrsandman, line 820: There should only be one reset in the whole lifetime of game

src/Game.cpp
```diff
@@ -826,12 +826,11 @@ void Game::LoadLandscape(const std::filesystem::path& path)
 	{
 		throw std::runtime_error("Could not find landscape " + path.generic_string());
 	}
-
-	_landIsland = std::make_unique<LandIsland>();
-	_landIsland->LoadFromFile(fixedName);
+	Locator::terrainSystem::emplace<LandIsland>();
```
bwrsandman, line 829: This should be done in initialize.

bwrsandman (commented):
The places LandIsland is created and destroyed is in the wrong place

src/Game.cpp
bwrsandman: I noticed some inconsistency in our code, our locator inits are in `Game::Initialize()`, `Locator::InitializeLevel()` or `InitializeGame()`

src/Game.h
```diff
@@ -179,22 +178,6 @@ class Game
...
-			throw std::runtime_error("Cannot get landscape before any are loaded");
-		}
-		return *_landIsland;
-	}
-	[[nodiscard]] LandIsland& GetLandIsland() const
-	{
-		if (_landIsland == nullptr)
-		{
-			throw std::runtime_error("Cannot get landscape before any are loaded");
-		}
```
bwrsandman, line 195: This error used to be thrown

src/LHScriptX/Script.cpp
```diff
@@ -119,7 +120,7 @@ ScriptCommandParameter GetParameter(Token& argument)
 				const auto z = std::strtof(floatEnd + 1, &floatEnd);
 				if (static_cast<size_t>(floatEnd - str.c_str()) == static_cast<size_t>(str.length()))
 				{
-					const auto& island = Game::Instance()->GetLandIsland();
+					const auto& island = Locator::terrainSystem::value();
```
bwrsandman, line 123: Now if script hasn't yet parsed `LOAD_LANDSCAPE`, this results in an assert fail.

bwrsandman (commented):
We're no longer catching errors in bad scripts and failing on an assert which is not ideal


Example script from `test_load_scene.cpp` `LoadScene::load_abode_no_landscape`
```
VERSION(2.300000)
CREATE_ABODE(0, "2224.63,2372.52", "CELTIC_ABODE_F", 11100, 1095, 0, 0)
```

src/ECS/Systems/Implementations/PathfindingSystem.cpp
```diff
@@ -27,6 +27,7 @@
 #include "ECS/Map.h"
 #include "ECS/Registry.h"
 #include "Game.h"
```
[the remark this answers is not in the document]
Sofistico: I just checked, but sadly not yet!
There are some calls that it does to some singletons inside the Game::Instance that will need to be decouple in another issue

src/Game.cpp
```diff
@@ -717,7 +717,7 @@ bool Game::Run()
 			    /*frameBuffer =*/nullptr,
 			    /*sky =*/*_sky,
 			    /*water =*/*_water,
-			    /*island =*/*_landIsland,
+			    /*island =*/Locator::terrainSystem::value(),
```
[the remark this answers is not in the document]
Sofistico: Fixed on the next commit

src/Locator.h
Sofistico: Fixed on the next commit

src/Game.cpp
Sofistico: fixed on the next commit

src/Game.cpp
Sofistico: I put the terrainSystem initilizer inside the Locator::InitializeLevel(), it seemed like the most appropriate place!

src/LHScriptX/Script.cpp
Sofistico: fixed on the next commit

src/Game.h
Sofistico: In regards to this i can only thing of adding a check somewhere to make sure that the landIsland is properly loaded, but i don't know any other way than checking every place the Game::Instance()->GetLandIsland() was called and checking if the Locator::terrainSystem::hasValue() and throwing the exception.
Do you have any other idea of what could be done?

src/3D/LandIslandInterface.h
```diff
@@ -0,0 +1,52 @@
...
+{
+class FrameBuffer;
+class Texture2D;
+} // namespace graphics
+class LandIslandInterface
+{
+public:
+	virtual void LoadFromFile(const std::filesystem::path& path) = 0;
+	[[nodiscard]] virtual float GetHeightAt(glm::vec2) const = 0;
+	[[nodiscard]] virtual const LandBlock* GetBlock(const glm::u8vec2& coordinates) const = 0;
```
bwrsandman, line 31: Unused so I wouldn't put it in the interface
```suggestion
```

src/3D/LandIsland.h
```diff
@@ -49,49 +49,50 @@ class LandIsland
...
 	~LandIsland();
 
-	void LoadFromFile(const std::filesystem::path& path);
+	void LoadFromFile(const std::filesystem::path& path) override;
 
-	[[nodiscard]] float GetHeightAt(glm::vec2) const;
-	[[nodiscard]] const LandBlock* GetBlock(const glm::u8vec2& coordinates) const;
-	[[nodiscard]] const lnd::LNDCell& GetCell(const glm::u16vec2& coordinates) const;
+	[[nodiscard]] float GetHeightAt(glm::vec2) const override;
+	[[nodiscard]] const LandBlock* GetBlock(const glm::u8vec2& coordinates) const override;
```
bwrsandman, line 55:
```suggestion
	[[nodiscard]] const LandBlock* GetBlock(const glm::u8vec2& coordinates) const;
```

src/3D/LandIsland.h
```diff
@@ -18,11 +18,11 @@
...
 
+#include "3D/LandIslandInterface.h"
 #include "LandBlock.h"
 
 namespace openblack
 {
-
 namespace ecs::systems
 {
 class DynamicsSystem;
```
bwrsandman, line 28: The `ecs::systems::DynamicsSystem` forward declare is unused.
There is already a `graphics::FrameBuffer` forward declare in the interface.
There header defining `lnd::LNDCountry` is already included.

src/3D/LandIslandInterface.h
```diff
@@ -0,0 +1,52 @@
...
+	[[nodiscard]] virtual const std::vector<lnd::LNDCountry>& GetCountries() const = 0;
+
+	[[nodiscard]] virtual const graphics::Texture2D& GetAlbedoArray() const = 0;
+	[[nodiscard]] virtual const graphics::Texture2D& GetBump() const = 0;
+	[[nodiscard]] virtual const graphics::Texture2D& GetHeightMap() const = 0;
+	[[nodiscard]] virtual const graphics::FrameBuffer& GetFootprintFramebuffer() const = 0;
+
+	virtual void GetIndexExtent(glm::u16vec2& extentMin, glm::u16vec2& extentMax) const = 0;
+	virtual void GetOrthoViewProj(glm::mat4& view, glm::mat4& proj) const = 0;
+	virtual void GetExtent(glm::vec2& extentMin, glm::vec2& extentMax) const = 0;
```
bwrsandman, line 49: You don't have to do this in this PR but I don't like this pattern and would prefer an extent struct being returned.

src/Game.h
bwrsandman: I don't think every call to GetLandIsland needs to be checked but there are certain places we know that it can happen

bwrsandman (changes requested):
There are still a lot of includes of the implementation file that should be replaced by the interface file.

```
src/3D/Camera.cpp:#include "3D/LandIsland.h"
src/3D/LandBlock.cpp:#include "LandIsland.h"
src/Debug/Gui.cpp:#include "LandIsland.h"
src/Debug/LandIsland.cpp:#include "LandIsland.h"
src/Debug/LandIsland.cpp:#include "3D/LandIsland.h"
src/Debug/MeshViewer.cpp:#include "3D/LandIsland.h"
src/Debug/PathFinding.cpp:#include "3D/LandIsland.h"
src/ECS/Systems/Implementations/CameraBookmarkSystem.cpp:#include "3D/LandIsland.h"
src/ECS/Systems/Implementations/DynamicsSystem.cpp:#include "3D/LandIsland.h"
src/ECS/Systems/Implementations/PathfindingSystem.cpp:#include "3D/LandIsland.h"
src/Game.cpp:#include "3D/LandIsland.h"
src/LHScriptX/FeatureScriptCommands.cpp:#include "3D/LandIsland.h"
src/LHScriptX/Script.cpp:#include "3D/LandIsland.h"
src/Renderer.cpp:#include "3D/LandIsland.h"
src/Serializer/FotFile.cpp:#include "3D/LandIsland.h"
```

Add this bit of code after the includes in LandIsland.h to prevent any issues like that happening again. Also define `LOCATOR_IMPLEMENTATIONS` in LandIsland.cpp


```c++
#if !defined(LOCATOR_IMPLEMENTATIONS)
#error "Locator interface implementations should only be included in Locator.cpp"
#endif
```

src/3D/LandIslandInterface.h
Sofistico: Hmm could you point me to what this could mean? I didn't find what the term extent struct means

bwrsandman: An extent is like a rectangle (in 2d, a rectangular cuboid in 3d) that is described by a minimum and maximum position.

```cpp
struct Extent2D
{
    glm::vec2 minima;
    glm::vec2 maxima;
}
```

Sofistico: Wound't a tuple work better? Since the ExtentIndex returns glm::u16vec2& and GetOrthoViewProj returns glm::mat4
Or should i create a struct for each?
`struct Extent2D
	{
		glm::vec2 minimum;
		glm::vec2 maximum;
	};
    struct IndexExtent
	{
	    glm::u16vec2 minimum;
	    glm::u16vec2 maximum;
	};
	struct OrthoView
	{
	    glm::mat4 proj;
	    glm::mat4 view;
	};`
Like this?

src/3D/LandIsland.h
Sofistico: Fixed on next commit

src/3D/LandIslandInterface.h
Sofistico: Fixed on next commit

src/3D/LandIsland.h
Sofistico: Fixed on next commit

src/3D/LandIslandInterface.h
bwrsandman: Problem with pairs and tuples is that members are unnamed and makes it hard to make sense of which data is which.

`GetOrthoViewProj` seems like it should be two getters.

Sofistico: Fixed on next commit!
please take a look later

src/3D/Extent2D.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#pragma once
+
+#include <glm/mat4x4.hpp>
```
bwrsandman, line 12: Not the right header for vec2.

src/3D/Extent2D.h
```diff
@@ -0,0 +1,26 @@
...
+{
+	glm::vec2 minimum;
+	glm::vec2 maximum;
+};
+struct IndexExtent
+{
+	glm::u16vec2 minimum;
+	glm::u16vec2 maximum;
+};
+} // namespace openblack
```
bwrsandman, line 26: Extent2D works well within `src/3D` but IndexExtent seems a little too specific to LandIsland.

What you could do is extend `glm` to have a template for all kinds of extents

```suggestion
#include <glm/detail/qualifier.hpp>

namespace openblack
{
template<glm::length_t L, typename T, glm::qualifier Q = glm::defaultp>
struct Extent
{
	glm::vec<L, T, Q> minimum;
	glm::vec<L, T, Q> maximum;
};

using Extent2 = Extent<2, float>;
using U16Extent2 = Extent<2, uint16_t>;

} // namespace openblack
```
The file would have to be renamed to Extent.

src/3D/LandBlock.h
```diff
@@ -72,8 +74,6 @@ class LandBlock
 	std::unique_ptr<btBvhTriangleMeshShape> _physicsMesh;
 	std::unique_ptr<btRigidBody> _rigidBody;
 
-	const bgfx::Memory* BuildVertexList(LandIsland& island);
-
-	friend LandIsland;
```
bwrsandman, line 77: Nice that we can get rid of this friend

src/3D/LandIsland.cpp
```diff
@@ -58,7 +62,7 @@ void LandIsland::LoadFromFile(const std::filesystem::path& path)
 	_landBlocks.resize(lndBlocks.size());
 	for (size_t i = 0; i < _landBlocks.size(); i++)
 	{
-		_landBlocks[i]._block = std::make_unique<lnd::LNDBlock>(lndBlocks[i]);
+		_landBlocks[i].GetLndBlock() = std::make_unique<lnd::LNDBlock>(lndBlocks[i]);
```
bwrsandman, line 65: Not a big fan of this. Should be using a SetLndBlock function.

src/Debug/LandIsland.cpp
```diff
@@ -41,11 +46,9 @@ void LandIsland::Draw(openblack::Game& game)
 
 	if (ImGui::TreeNodeEx("Height Map", ImGuiTreeNodeFlags_DefaultOpen))
 	{
-		glm::u16vec2 extentMin;
-		glm::u16vec2 extentMax;
-		landIsland.GetIndexExtent(extentMin, extentMax);
-		const auto extentSize = extentMax - extentMin;
-		const auto dim = static_cast<uint16_t>(openblack::LandIsland::k_CellCount) * extentSize;
+		auto indexExtent = landIsland.GetIndexExtent();
```
bwrsandman, line 49:
```suggestion
		const auto indexExtent = landIsland.GetIndexExtent();
```

bwrsandman (changes requested):

src/3D/Extent2D.h
Sofistico: Fixed on next commit!

src/3D/LandIsland.cpp
Sofistico: Fixed on next commit

src/3D/LandBlock.h
Sofistico: Friend is a weird stuff in cpp....
Wasn't expecting such a weird thing!

src/Debug/LandIsland.cpp
Sofistico: Fixed on next commit

bwrsandman (approved):
I'm really happy how this turned out. I'll add some header optimization just for the sake of speeding up compilation and will merge once the tests for that pass.

raffclar:
This is looking great

src/LHScriptX/Script.cpp
```diff
@@ -119,7 +120,12 @@ ScriptCommandParameter GetParameter(Token& argument)
 				const auto z = std::strtof(floatEnd + 1, &floatEnd);
 				if (static_cast<size_t>(floatEnd - str.c_str()) == static_cast<size_t>(str.length()))
 				{
-					const auto& island = Game::Instance()->GetLandIsland();
+					if (!Locator::terrainSystem::has_value())
+					{
+						throw std::runtime_error("Cannot get landscape before any are loaded");
```
raffclar, line 125: Repeating this conditional throw in lots of places smells. We don't check other locator patterns before usage. I think letting the actual error from accessing the value throw is better

raffclar: @bwrsandman Thoughts?

raffclar (changes requested):

src/LHScriptX/Script.cpp
bwrsandman: This one is a valid error. We actually have a test to catch a `runtime_error`, not an assert: `LoadScene, load_abode_no_landscape`. This happens when you don't call `LOAD_LANDSCAPE` early enough. This is definitely something we want to check in this instance.

The others maybe not. I can remove them.

raffclar (approved):

PR #620, opened 2023-07-02
Implement Audio

raffclar:
#273

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,13 @@ if (TARGET bgfx::bgfx AND TARGET bgfx::bx)
   target_link_libraries(bgfx::bgfx INTERFACE bgfx::bx)
 endif ()
 
+find_path(
+  DRLIBS_INCLUDE_DIRS REQUIRED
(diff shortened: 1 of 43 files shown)


bwrsandman:
Fixes done in #621

bwrsandman:
The cross compile issues are due to https://github.com/kcat/openal-soft/issues/871

.clang-format-ignore
```diff
@@ -6,3 +6,5 @@ src/Common/stb_image_write.h
 test/mobile_wall_hug/json.hpp
 externals/imgui_user/imgui_impl_sdl.h
 externals/imgui_user/imgui_impl_sdl.cpp
+externals/dr_libs/dr_mp3.h
```
raffclar, line 9: Why not use vcpkg?

bwrsandman: Because the package is not common on linux platforms and ti would break the system-deps

bwrsandman:
Putting to draft until the git module change commit is remove (i.e. vcpkg's openal-soft pr gets merged).

bwrsandman (approved):

PR #621, opened 2023-07-03
Audio system 3

bwrsandman:
Clean-up of #620 

* Squashed fixup commits
* Removed unnecessary changes (whitespace in json and different permissions on python files)
* Move Audio components to src/Audio
* Add no-op audio manager in case system has no audio.
* Removed the `Sound` because data is already available in `PackFile`. You get the audio file header from one call and the bytes from another. It's much more in line with other functions in packfile.
* Removed SadBankInfo because it is not used anywhere so I don't have any way to validate.
    * Would be good to add that back but in a separate PR
* dr_libs are in `externals/` so that non-vcpkg versions will work

### DIFF
--- a/src/Audio/AudioPlayer.cpp
+++ b/src/Audio/AudioPlayer.cpp
@@ -1,0 +1,181 @@
+/******************************************************************************
+ * Copyright (c) 2018-2023 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#include "AudioPlayer.h"
+
+#include <random>
+
+#include "AlCheck.h"
+
+using namespace openblack::audio;
+
+void AudioPlayer::Initialize()
+{
+	alCheckCall(alcMakeContextCurrent(_context.get()));
+}
+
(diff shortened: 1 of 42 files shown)


src/Audio/AudioPlayer.cpp
```diff
@@ -0,0 +1,242 @@
...
+                                      [[maybe_unused]] int length) noexcept
+{
+	enum class LogLevel
+	{
+		Disable,
+		Error,
+		Warning,
+		Trace
+	};
+	switch (static_cast<LogLevel>(level))
```
bwrsandman, line 148: Wrong. It will be E,W,T

src/Audio/AudioPlayer.cpp
```diff
@@ -0,0 +1,242 @@
...
+		alsoftSetLogCallback(AudioPlayerAlLogger, nullptr);
+	}
+	else
+	{
+		SPDLOG_LOGGER_WARN(spdlog::get("audio"), "Could not set openal logging callback");
+		SPDLOG_LOGGER_WARN(spdlog::get("audio"), "Falling back to tracing directly in openal");
+#if defined(_MSC_VER)
+		_putenv_s("ALSOFT_LOGLEVEL", "3");
+#else
+		setenv("ALSOFT_LOGLEVEL", "3", 1);
```
bwrsandman, line 182: Get level from audio logger level

bwrsandman:
moved these changes to #620

PR #624, opened 2023-07-06
Faster Transformation matrix for entities

FalseIlyu:
When preparing transformation of entities for the draw call we currently create a 4x4 identity matrix translate and multiply it by the rotation as a 4x4.
Here we build a 4x4 matrix with the rotation of the transform then translate it by `position x transform` and use a smaller multiplication.

| Debug Master | Debug Fast |
|----------------|----------------|
|masterDebug|fastermatrix|
| Release Master | Release Fast |
|releasedMaster|realeaseFast|


bwrsandman (approved):
LGTM. Just have to test it later.

PR #629, opened 2023-07-10
Decoupling entitiesMap from Game

### LINKED ISSUE #626: Decouple Entity Registry and Map from Game.h
Part of a push to remove the Game singleton.

Use the locator pattern to get access to the LandIsland handle.

Example usage
```c++
// Current
#include "Game.h"
#include "ECS/Map.h"
#include "ECS/Registry.h"
const auto& map = Game::Instance()->GetEntityMap();
auto& registry = Game::Instance()->GetEntityRegistry();

// Better
#include "Locator.h"
#include "ECS/Map.h"
#include "ECS/Registry.h"
const auto& map = Locator::entityMap::value();
auto& registry = Locator::entityRegistry::value();
```

Since `Registry` is so heavily templated, it might not be possible to create and interface for it. However `Map` can be.

FalseIlyu:
Creation of a MapInterface and MapProduction class used through the Locator Class.
Closes #626.
Review welcome but i keep this PR in draft to keep the possibility to add the Registry here.

Edit : Added entitiesRegistry, be aware that i still need to look into the file dependant upon entitiesRegistry and entitiesMap to see if i can cull inclusion of Game.h.


src/Debug/PathFinding.cpp
```diff
@@ -37,6 +37,8 @@ PathFinding::PathFinding()
 
 void PathFinding::Draw(Game& game)
 {
+	// TODO(falseilyu): #626 Get a better way to deal with unused parameter warning.
+	std::ignore = game;
```
bwrsandman, line 41: [[maybe_unused]] in front of the parameter.

FalseIlyu: Thanks

src/ECS/Map.cpp
```diff
@@ -19,96 +19,26 @@
 #include "ECS/Components/Mobile.h"
 #include "ECS/Components/Transform.h"
 #include "ECS/Registry.h"
-#include "Game.h"
+#include "Locator.h"
```
bwrsandman, line 22: Lots of unused headers in here

src/ECS/MapProduction.h
```diff
@@ -0,0 +1,34 @@
...
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#pragma once
+
+#include "Map.h"
+
```
bwrsandman, line 13:
```suggestion
#if !defined(LOCATOR_IMPLEMENTATIONS)
#warning "Locator interface implementations should only be included in Locator.cpp, use interface instead."
#endif
```

src/ECS/MapProduction.cpp
```diff
@@ -0,0 +1,95 @@
+/******************************************************************************
+ * Copyright (c) 2018-2023 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
```
bwrsandman, line 9:
```suggestion

#define LOCATOR_IMPLEMENTATIONS

```

src/ECS/Registry.h
```diff
@@ -147,11 +135,12 @@ class Registry
 	template <typename... Components>
 	[[nodiscard]] decltype(auto) Size() const
 	{
-		return _registry.view<Components...>().size();
+		return GetRegistry().view<Components...>().size();
 	}
 
 private:
-	entt::registry _registry;
+	[[nodiscard]] virtual entt::registry& GetRegistry() = 0;
```
bwrsandman, line 142: Moving to a virtual function adds an indirection and a guaranteed cache miss.
As mentioned in #626 it might not be necessary to make this virtual.

As I see it, this doesn't improve compilation time: There's only `_registry` that gets hidden and entt.hpp is still included in the interface.
This class, and `_registry especially`, is a hotspot for performance so I don't think making it an abstract base class is the right move.

bwrsandman (changes requested):
Very good work. This touched a lot of file and it looks much cleaner now.

I have some worries about `_registry` though.

Ran some manual tests and everything looks okay.

bwrsandman (approved):

PR #635, opened 2023-07-11
use stb from system/vcpkg

bwrsandman:


bwrsandman:
stb is in vcpkg and a community package in linux, however the linux version does not have cmake configuration.

PR #643, opened 2023-07-22
Temple misc fixes

### LINKED ISSUE #642: Camera at nan position when entering the temple
When entering the temple from certain position/angle the camera position become a vector of nan. Once it happen the camera always end up there no matter if you move the camera when leaving.
Tested on windows with Direct3D12 backend

nanPosition
nan
[nanLog.txt](https://github.com/openblack/openblack/files/12136825/nanLog.txt)

FalseIlyu:
Fixes #642 by moving the camera from the temple position by one on the Y axis so no zero vector appears in calculation.
Fix a bug where the Desactivation of the interior do not lead to the removal of interior entity.

Add a _templeRotation so the camera rotation is set when entering and aligned with the temple.

### DIFF
--- a/src/3D/TempleInterior.cpp
+++ b/src/3D/TempleInterior.cpp
@@ -108,7 +108,7 @@ void TempleInterior::Deactivate()
 	auto& config = Game::Instance()->GetConfig();
 	config.drawIsland = true;
 	config.drawWater = true;
-	registry.Each<const TempleInterior>([&registry](const entt::entity entity, auto&&...) { registry.Destroy(entity); });
+	registry.Each<const TempleInteriorPart>([&registry](const entt::entity entity, auto&&...) { registry.Destroy(entity); });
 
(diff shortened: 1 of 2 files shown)


raffclar:
Checking

src/3D/TempleInterior.cpp
```diff
@@ -108,7 +111,7 @@ void TempleInterior::Deactivate()
 	auto& config = Game::Instance()->GetConfig();
 	config.drawIsland = true;
 	config.drawWater = true;
-	registry.Each<const TempleInterior>([&registry](const entt::entity entity, auto&&...) { registry.Destroy(entity); });
+	registry.Each<const TempleInteriorPart>([&registry](const entt::entity entity, auto&&...) { registry.Destroy(entity); });
```
raffclar, line 114: Nice

raffclar (approved):

Issue #646, opened 2023-07-29
Create a separate renderering of interior temple.


FalseIlyu:
After diagnosis in #641, the conclusion is that the problem is that we just use too much memory.

Here from most important to less important
- [x] Create a specific renderer only for the temple interior.
- - [x] This renderer need only to render Interior part currently used by the game.
- - [x] The normal renderer need to not render interior part.

FalseIlyu:
Removing specification related to the door stuff (and the pause), this stuff will be treated at a latter date.

FalseIlyu:
Implementing by #647

PR #647, opened 2023-07-29
Separate the rendering of the Temple

FalseIlyu:
Try to implement #646, this try to leverage the locator and renderingSystem and switch the rendering system when activating the temple.
The normal rendering system filter out temple interior entity, the temple rendering system filter out all that is not a temple interior entity.

- [x] Create a specific renderer only for the temple interior.
- - [x] This renderer need only to render Interior part currently used by the game.
- - [x] The normal renderer need to not render interior part.

EDIT
Here is an example of the temple without island entities rendered.
templeAlone
Be aware, dynamic loading of the temple not being implemented entering the temple is still not possible on vulkan.


raffclar (approved):
Really nice

raffclar:
@FalseIlyu ok to merge?

FalseIlyu:
> @FalseIlyu ok to merge?

If there are no request it is ok for me.

PR #653, opened 2023-08-05
Set blockPosition through InstanceDataBuffer

FalseIlyu:
Use an uniform array in LandIsland to feed the u_blockPositionAndSize uniform as instance data inside vs_terrain.sc.


bwrsandman (commented):
As stated previously, it's a bit strange to use instance buffers if you're not using instanced drawing

FalseIlyu:
> As stated previously, it's a bit strange to use instance buffers if you're not using instanced drawing

Setting the uniform in the loop is slower.
I know it's not ideal but i think it's worse to have data that vary with each call as uniforms.

Issue #617, opened 2023-06-30
Decouple `lhscriptx::Script()` from Game interface and clean-up.


bwrsandman:
To follow #616 

With the use of locators, we no longer have any use to keep a pointer to game in `lhscriptx::Script()`.

Additionally, there is a lot of unused code and clean-up that can be done:
- [x] `lhscriptx::NotImplemented`
- [x] Commented out functions of `lhscriptx::Script`
- [x] `lhscriptx::Script` constructor can be set to default (in cpp file) if `_token` is set in declaration.
- [x] `lhscriptx::ScriptContext::GetParameters`
- [x] `lhscriptx::ScriptContext` can probably be removed.

PR #654, opened 2023-08-15
Lh script decoupling

FalseIlyu:
Implement #617 

* [x]  removal of `lhscriptx::NotImplemented`
* [x]  removal of  commented out functions of `lhscriptx::Script`
* [x]  `lhscriptx::Script` constructor set to default (in cpp file), `_token` is set in declaration.
* [x]  removal of  `lhscriptx::ScriptCommandContext::GetParameters`
* [x]  removal of  `lhscriptx::ScriptCommandContext`.

The last point will need a further look into how `lhscriptx::ScriptCommandContext` is used as it should be removable but will need a look as to where and it is used in the different part of lhscriptx.

### DIFF
--- a/src/Debug/Console.cpp
+++ b/src/Debug/Console.cpp
@@ -365,7 +365,7 @@ void Console::Draw(Game& game)
 
 		if (s[0] != 0)
 		{
-			ExecCommand(s, game);
+			ExecCommand(s);
 		}
 
 		_inputBuffer[0] = '\0';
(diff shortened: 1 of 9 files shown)


bwrsandman (changes requested):

sandy-carter-unity (approved):

sandy-carter-unity (changes requested):
You can now remove `class Game` forward declares from Script.h and `CommandSignature.h`

bwrsandman (changes requested):
You can now remove `class Game` forward declares from `Script.h` and `CommandSignature.h`

PR #658, opened 2023-08-20
Fix camera projection reciprocity

bwrsandman:
Camera Deproject and project were not reciprocal due to yflip in de-project but not in project.

DeprojectScreenToWorld:
* Return Ray structure instead of relying on side effects
* Make immutable
* Use u16 for type safety

ProjectWorldToScreen:
* Return optional instead of relying on side effects
* Make use of Extent

Added unit test for camera.

This is foundational work for the new camera model.


test/test_camera.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+	const auto screenSize = glm::u16vec2 {k_Width, k_Height};
+	const auto result = _camera->ProjectWorldToScreen(glm::vec3(0.0f, 0.0f, 1.0f), U16Extent2 {{0, 0}, screenSize});
+	ASSERT_TRUE(result.has_value());
+	ASSERT_EQ(static_cast<glm::u16vec2>(glm::xy(result.value())), glm::u16vec2(k_Width / 2, k_Height / 2));
+}
+
+TEST_F(TestCamera, project_deproject)
+{
+	auto screenPoints = std::array {
```
bwrsandman, line 45: make const

test/test_camera.cpp
```diff
@@ -0,0 +1,57 @@
...
+	    glm::u16vec2 {k_Width / 2, k_Height / 2},
+	    glm::u16vec2 {k_Width / 4, k_Height / 4},
+	};
+	for (auto screenPoint : screenPoints)
```
bwrsandman, line 49: make const ref

test/mobile_wall_hug/test_mobile_wall_hug.cpp
```diff
@@ -25,19 +25,11 @@
 #include <spdlog/sinks/stdout_color_sinks.h>
 
 #include "json.hpp"
+#include "third_party/glm_helpers.h"
```
bwrsandman, line 28: remove `<glm/gtx/string_cast.hpp>`

test/test_camera.cpp
```diff
@@ -0,0 +1,63 @@
...
+		const auto screenSize = glm::u16vec2 {k_Width, k_Height};
+		const auto ray = _camera->DeprojectScreenToWorld(screenPoint, screenSize);
+		const auto result = _camera->ProjectWorldToScreen(ray.origin, U16Extent2 {{0, 0}, screenSize});
+		ASSERT_TRUE(result.has_value());
+		if (result.has_value())
+		{
+			ASSERT_EQ(static_cast<glm::u16vec2>(glm::xy(result.value())), screenPoint);
+		}
+	}
+}
```
bwrsandman, line 63: Add a test for a point at the same position as the camera.

raffclar (approved):

Issue #213, opened 2020-02-08
Beach runoff into ocean glitching


sylcai:
Under certain camera angles the beaches completely vanish
![image](https://user-images.githubusercontent.com/5037244/74087962-78770200-4a91-11ea-9649-37816842f91c.png)
![image](https://user-images.githubusercontent.com/5037244/74087969-80cf3d00-4a91-11ea-8e45-fee4368510c9.png)
Blinking in and out of existence

sylcai:
![beachgone](https://user-images.githubusercontent.com/5037244/74088053-5af66800-4a92-11ea-82dd-e331528e9dc6.gif)

PR #665, opened 2023-11-20
Reversing the depth

FalseIlyu:
Reversing (1 for the near plane, 0 for the far plane) the Depth value in order to avoid Z-Fighting as observed in #213  (which will be solved by this PR).

Changes the camera perspective matrix and the BGFX state in the renderer.
Superficial check were done on :
- [x] OpenGL
- [x] OpenGLES
- [x] Vulkan
- [x] Direct3D11
- [x] Direct3D12
- [x] Metal
- ~Gnm~
- ~Nvn~
- ~Noop~

Example of #213
213notsolargedist
Same viewpoint now
screenshot

TBD : Checking that the camera does behave the same and that the math are correct.

### DIFF
--- a/src/3D/Camera.cpp
+++ b/src/3D/Camera.cpp
@@ -134,11 +134,11 @@ void Camera::ResetVelocities()
 void Camera::SetProjectionMatrixPerspective(float xFov, float aspect, float nearClip, float farClip)
 {
 	const float yFov = (glm::atan(glm::tan(glm::radians(xFov) / 2.0f)) / aspect) * 2.0f;
-	const float h = -1.0f / tan(yFov * 0.5f);
-	const float w = -h / aspect;
+	const float h = 1.0f / glm::tan(yFov * 0.5f);
+	const float w = h / aspect;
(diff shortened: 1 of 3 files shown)


bwrsandman:
We don't support Gnm Nvn as they're NDA platforms for which I don't have licenses for this project.
I can help test Metal and I think you're good for the no-op renderer.

FalseIlyu:
Investigation ongoing, but i manage to get this on master branch so it is not something introduced by the change in perspective
screenshot

FalseIlyu:
So tracking down what happens, we arrive at RayCastClosestHit with a normal {0.0, -1.0, 0.0} which is not guarded against in glm::orientation but will return a rotation full of nan.
This didn't happen but can now happen high value in the x rotation.

bwrsandman:
I saw this issue during my camera model refactoring. I'll try to find my solution

FalseIlyu:
Finally got around to test the camera behavior and look where changing the depth impact the camera behavior. The camera should now behave like in the master branch.

bwrsandman (approved):
Validated on Metal

PR #671, opened 2023-12-29
packtool: Fix view-bytes option

bwrsandman:


FalseIlyu:
Tried with `D:\openblack\cmake-build-presets\msvc-17-vcpkg\bin\Debug\packtool.exe -B 'C:\...\Data\AllMeshes.g3d' -b "MESHES"`. Doesn't work on master, works with the fix, nothing more to say.

PR #672, opened 2023-12-29
Add support for B&W files from v1.00

### LINKED ISSUE #281: Cannot load gamedata
So I've been trying to get openblack to load the gamedata.

I keep running into an error `Failed to open Scripts/info.dat: Info block size does not match that of GInfo: 580710 != 582022`

I first tried unpacking the data through unshield, which seemed to work reasonably enough except for some rearranging (#280).
I ran into the above error though.

So I set up a VM with WinXP and fully instaled the game there and moving it back to my linux system. I still ran into the same error though.

Do I need some other version of the gamedata? Should it be with the creature isle expansion? What am I doing wrong?

bwrsandman:
Fixes #281

### DIFF
--- a/src/InfoConstants.h
+++ b/src/InfoConstants.h
@@ -22,6 +22,7 @@
 namespace openblack
 {
 #pragma pack(push, 1)
+
 struct GTribeInfo
 {
 	Tribe tribeType;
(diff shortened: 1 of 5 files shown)


handsomematt (approved):
Nice use of namespaces

src/InfoConstants.h
```diff
@@ -1801,6 +1741,68 @@ struct GGroundInfo: GObjectInfo
 	MeshId editorMesh;
 };
 
+namespace v100 // original relase 1.0
```
bwrsandman, line 1744:
```suggestion
namespace v100 // original release 1.0
```

PR #673, opened 2023-12-29
filesystem: Move pre-compiler checks to interface

bwrsandman:
This allows for more interfaces to use custom read functions than android.
Such a usecase would be single ISO reads.

### DIFF
--- a/src/3D/LandIsland.cpp
+++ b/src/3D/LandIsland.cpp
@@ -48,12 +48,14 @@ void LandIsland::LoadFromFile(const std::filesystem::path& path)
 
 	try
 	{
-#if __ANDROID__
-		//  Android has a complicated permissions API, must call java code to read contents.
-		lnd.Open(Locator::filesystem::value().ReadAll(path));
-#else
-		lnd.Open(path);
-#endif
+		if (Locator::filesystem::value().PreferBuffer())
+		{
+			lnd.Open(Locator::filesystem::value().ReadAll(path));
+		}
+		else
+		{
+			lnd.Open(path);
+		}
 	}
 	catch (std::runtime_error& err)
 	{
--- a/src/3D/Sky.cpp
+++ b/src/3D/Sky.cpp
@@ -33,13 +33,15 @@ Sky::Sky()
 
 	// load in the mesh
 	_model = std::make_unique<L3DMesh>("Sky");
-#if __ANDROID__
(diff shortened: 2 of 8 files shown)


FalseIlyu:
Tested on a windows computer \:
Normally launching the game works without problem, i tried setting the default filesystem preferedBuffer to true and it seems to cause no issue too and everything seems right to me.


However i wonder if we could not have a way to forgo the check and change this
```
if (Locator::filesystem::value().PreferBuffer())
{
	lnd.Open(Locator::filesystem::value().ReadAll(path));
}
else
{
	lnd.Open(path);
}
```
Into and just hide the whether the file is read from a buffer or from the filesystem.
```
lnd.Open(path);
```

bwrsandman:
I had the same thought to reduce duplicate code but couldn't find a satisfying implementation.

The issue is that a static lib like `lnd` would not have details of custom filesystem read implementations if they took charge of reading.

And since I want components to be independent of each other, I don't want to have some monster reader common class.

FalseIlyu:
Wouldn't going the other way around and having things written as
```
lnd.Open(Locator::filesystem::value().ReadAll(path));
```
Maybe not exactly like this something of the effect of the filesystem is in charge of getting the data to the reader and then reader do its job.

bwrsandman:
So just always reading the whole file and using the buffer implementation.
I wonder if there are performance implications to consider such as the cost of allocation or read time.
That being said SSDs are fast nowadays.

It's worth considering and would be a non trivial change in the way we interact with our files.

FalseIlyu:
```
void PackFile::Open(const std::filesystem::path& file)
{
	_filename = file;

	std::ifstream stream(_filename, std::ios::binary);

	if (!stream.is_open())
	{
		Fail("Could not open file.");
	}

	ReadFile(stream);
}

void PackFile::Open(const std::vector<uint8_t>& buffer)
{
	assert(!_isLoaded);

	imemstream stream(reinterpret_cast<const char*>(buffer.data()), buffer.size() * sizeof(buffer[0]));

	// File name set to "buffer" when file is load from a buffer
	// Impact code using L3DFile::GetFilename method
	_filename = std::filesystem::path("buffer");

	ReadFile(stream);
}
```
Seeing as the Open method of many file reader is mostly setting up a stream for ReadFile couldn't we just make ReadFile accessible to LND and charge FileSystem with the job of setting up the stream?

bwrsandman:
Superseded by #706

PR #678, opened 2023-12-31
Update vcpkg

raffclar:


bwrsandman:
Duplicate of #683 which is merged now

Issue #670, opened 2023-12-13
Use a more explicit function to check if game path is valid.


bwrsandman:
At the time of writing, we check the existence and validity of game install path with `GetGamePath().empty()`:
```c++
	fileSystem.SetGamePath(_gamePath);

	if (fileSystem.GetGamePath().empty())
	{
		// no key, don't guess, let the user know to set the command param
		SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, "Game Path missing",
		                         "Game path was not supplied, use the -g "
		                         "command parameter to set it.",
		                         nullptr);
		SPDLOG_LOGGER_ERROR(spdlog::get("game"), "Failed to find the GameDir.");
		return false;
	}
```

This is a bad pattern and doesn't lend itself well to different implementations.

A better pattern would be to use a function to check the validity of the path either before or after setting.

An even better pattern would be to score and fall back on different filesystem implementations based on the given path.

PR #686, opened 2024-01-06
Filesystem: Add `IsPathValid()`

Afr0:
Fix for #670

Added `IsPathValid()` function to DefaultFileSystem and AndroidFileSystem. And use to validate game path at launch.

Compiled on win and Android, tested on Win.


src/FileSystem/DefaultFileSystem.cpp
```diff
@@ -72,6 +74,22 @@ std::filesystem::path DefaultFileSystem::FindPath(const std::filesystem::path& p
...
+	if (path.empty())
+		return false;
+
+	std::ifstream file(path);
+
+	if (file.is_open())
+	{
+		file.close();
+		return true;
+	}
```
bwrsandman, line 88: `path` is most likely going to be a directory (though with #674 it will have the option to be a file) so this won't work.

Have a look at [`std::filesystem::exists`](https://en.cppreference.com/w/cpp/filesystem/exists) which is most likely what you want here.

bwrsandman (changes requested):

src/FileSystem/DefaultFileSystem.cpp
Afr0: @bwrsandman Please see updated fix :)

bwrsandman: Seems like you're still opening the directory like if it was a file

bwrsandman:
Converting to draft until you can get it running without crashing

Afr0:
@bwrsandman Please see latest fix, should build on Android as well now.

src/FileSystem/AndroidFileSystem.cpp
```diff
@@ -50,6 +52,22 @@ std::filesystem::path AndroidFileSystem::FindPath(const std::filesystem::path& p
 	return path;
 }
 
+bool DefaultFileSystem::IsPathValid(const std::filesystem::path& path)
```
bwrsandman, line 55:
```suggestion
bool AndroidFileSystem::IsPathValid(const std::filesystem::path& path)
```

src/FileSystem/AndroidFileSystem.cpp
```diff
@@ -50,6 +52,22 @@ std::filesystem::path AndroidFileSystem::FindPath(const std::filesystem::path& p
...
+
+	std::ifstream file(path);
+
+	if (std::filesystem::exists(path))
+	{
+		file.close();
+		return true;
+	}
+
+	return false;
```
bwrsandman, line 68: This is not how the android filesystem works. Since android sandboxes the filesystem, you need to call into java code to use the android API.
See the rest of the code.

src/FileSystem/DefaultFileSystem.cpp
```diff
@@ -72,6 +74,22 @@ std::filesystem::path DefaultFileSystem::FindPath(const std::filesystem::path& p
 	throw std::runtime_error("File " + path.string() + " not found");
 }
 
+bool DefaultFileSystem::IsPathValid(const std::filesystem::path& path)
+{
+	if (path.empty())
+		return false;
+
+	std::ifstream file(path);
```
bwrsandman, line 82: This will fail, as discussed previously.

bwrsandman (changes requested):
There are major issues with this PR.
Please have a look and **test locally**.

Afr0:
@bwrsandman It seems to be working now, but the tests are failing because now they need the game installed for the test to pass...

bwrsandman:
The game is installed in the test environment but your changes lead the test to think that they aren't.
That's a bug in your PR.

> [Game.cpp:485] Failed to find the GameDir.

Afr0:
@bwrsandman Everything works now, please review.

src/FileSystem/AndroidFileSystem.cpp
```diff
@@ -50,6 +55,26 @@ std::filesystem::path AndroidFileSystem::FindPath(const std::filesystem::path& p
...
+	}
+
+	if (!std::filesystem::is_directory(path))
+	{
+		return false;
+	}
+
+	return true;
+}
+
```
bwrsandman, line 77: This won't work. It needs to use the java api

bwrsandman (changes requested):

src/FileSystem/AndroidFileSystem.cpp
Afr0: @bwrsandman Ok, on it. How come it passed the tests?

bwrsandman: You will have to setup android studio

FalseIlyu: > @bwrsandman Ok, on it. How come it passed the tests?

It passes the tests because we don't have runtime test on android only compilation test so what you wrote is correct c++ but due to Android ~PITA~ sandboxed filesystem using std::filesystem for android will never work when running the code you need to using android Java API.

This means anything you need the filesystem to do needs to be in a java method, preferably a static one inside the FileSystemInterop class we already have this class call the android API to do what you need in the filesystem. Then in the c++ side you need to use the JNI (a context is already provided to you inside the AndroidFileSystem class) to call this method from the JVM an use it's result in the app.

Afr0:
@bwrsandman Check it out

bwrsandman:
@Afr0 Do you mind editing the title and description of the PR with a description and summary of the change. Things you tested, things I should know about the change. Why things were done the way they were.

https://www.pullrequest.com/blog/writing-a-great-pull-request-description/

src/FileSystem/DefaultFileSystem.cpp
```diff
@@ -72,6 +77,26 @@ std::filesystem::path DefaultFileSystem::FindPath(const std::filesystem::path& p
...
+	{
+		return false;
+	}
+
+	if (!Exists(path))
+	{
+		return false;
+	}
+
+	if (!std::filesystem::is_directory(path))
```
bwrsandman, line 92: There will be confusion on this function call if we rule out directories.
Either
1. Remove this restriction
2. Change the name so it's clear we're checking only directories
3. Add an option to the interface for the restriction

src/FileSystem/AndroidFileSystem.cpp
```diff
@@ -49,6 +54,35 @@ std::filesystem::path AndroidFileSystem::FindPath(const std::filesystem::path& p
...
+	{
+		spdlog::error("Failed to find method: getDirectoryFromPath");
+		return false;
+	}
+
+	jobject juri = _jniEnv->CallStaticObjectMethod(_jniInteropClass, midGetDirectoryFromPath, _jniActivity, jgamePath, jpath);
+
+	bool isValid = juri != nullptr;
+
+	if (juri != nullptr)
```
bwrsandman, line 75:
```suggestion
	if (isValid)
```

src/FileSystem/DefaultFileSystem.cpp
```diff
@@ -72,6 +77,26 @@ std::filesystem::path DefaultFileSystem::FindPath(const std::filesystem::path& p
 	throw std::runtime_error("File " + path.string() + " not found");
 }
 
+bool DefaultFileSystem::IsPathValid(const std::filesystem::path& path)
```
bwrsandman, line 80: The way this is implemented requires the game path to be already set (`Exists(path)`). This is a problem because this needs to be able to be called before setting the game path.

The whole point of this function is to verify is a path is valid so we can then call the set to game path.

src/Game.cpp
```diff
@@ -469,7 +469,7 @@ bool Game::Initialize()
 
 	fileSystem.SetGamePath(_gamePath);
```
bwrsandman, line 470: This call should be moved to after the check.

bwrsandman (changes requested):

bwrsandman (approved):

PR #701, opened 2024-03-26
Ci tests/gradle report upload

FalseIlyu:
Required for #691 
Upload the report from gradle lint as an artefact in case of failure of the lint check.

As it is a bit hastily/lazily made this PR should be squashed.


FalseIlyu:
libsdl java sources were excluded from the linting scope

bwrsandman (approved):

Issue #712, opened 2024-04-22
Remove the `OldCameraModel`


bwrsandman:
This model has been replaced and should be removed.

PR #714, opened 2024-04-22
New Camera model

bwrsandman:
This is the result of over a year of research and work.

Depends on: 
- [x] #586 
- [x] #722 

This PR splits the camera data and the camera behaviour into `Camera` and `CameraModel`.
This allows for switching of camera models for rails or for temple navigation.
The camera is now controlled by two points in space `origin` and `focus`, camera orientation comes from the look-at vector that these two points create.
These points are controlled by the camera model logic and are smoothed out on the camera itself using the interpolators from #698.
In order to have the closest possible experience to vanilla, I recorded, using a mod of my making, a series of interactions for which I took the values of the in-game camera controller.
The old camera behaviour is put in a camera model named `OldCameraModel`. I did my best to preserve its functionality but I can't promise that it works exactly as it did before. It will eventually be removed (#712).

These are the interactions:

### Cartesian

Regular movement with the arrow keys

[Screencast from 2024-04-21 23-54-54.webm](https://github.com/openblack/openblack/assets/1013356/140ba2fd-03bc-496e-b87d-a3a48a8093a4)

### Polar

Keyboard rotation when pressing Shift and the arrow keys

[Screencast from 2024-04-21 23-56-35.webm](https://github.com/openblack/openblack/assets/1013356/8a44fbc4-6b10-4b9b-8eb8-9f4907dea3f4)

### Ctrl Zooming

Pressing Ctrl and the up and down arrow keys zooms. The left and right arrow keys do the same as shift

[Screencast from 2024-04-21 23-59-14.webm](https://github.com/openblack/openblack/assets/1013356/0b05e33d-fdfc-4b09-a906-f16edc2b6a56)

### Mouse wheel zooming

Regular zooming while rolling the scroll wheel

[Screencast from 2024-04-21 23-59-43.webm](https://github.com/openblack/openblack/assets/1013356/94bd41e6-58b1-4a91-adcf-189d8a30dd9c)
zooming

### Two button zoom

Zooming while holding both the left and right button down and moving forward and backwards with the mouse

[Screencast from 2024-04-22 00-00-50.webm](https://github.com/openblack/openblack/assets/1013356/aa89ee88-578d-4961-a2b6-e514fede9319)

### Grabbing Landscape

When you click and drag on the land or water you move inversely to the hand movement

[Screencast from 2024-04-22 00-11-52.webm](https://github.com/openblack/openblack/assets/1013356/7bbe8754-b41e-4821-bca0-ee8338f47ec1)

### Arcballing

When you middle click on the mouse and drag, the camera will move around that point in an arcball. This is also the way you move when viewing creature fights.

I find it hard to engage in vanilla and it's therefore as difficult in the re-implementation.

[Screencast from 2024-04-22 00-15-22.webm](https://github.com/openblack/openblack/assets/1013356/ff70c61d-b082-4aa3-9830-8fecbe0f71c2)

### Fly-to

This is whenever you double click on land or press and shortcut button such as bookmarks or the temple

[Screencast from 2024-04-22 00-13-30.webm](https://github.com/openblack/openblack/assets/1013356/3067efe4-2771-46bb-86df-4ce0fb513913)


FalseIlyu:
During a fly-by if you try to arcball, the game crashes from a bad optional access.
In more detail it seems to always occurs when you try to arc-ball right after starting a fly-by, if you do it toward the end of the fly-by it doesn't crash.

bwrsandman:
@FalseIlyu I am unable to repro the crash on Linux using the vcpkg multi ninja config in debug.

How I tried: I found a place where middle button rotates the camera, double click and immediately middle click. I tried this for both near and far fly-to.

Maybe it has to do with you trying this on windows? Could you show me a stack trace?

FalseIlyu:
Here is the trace, sorry i didn't get it earlier
```
_CxxThrowException 0x00007ffbe2cab640
std::_Throw_bad_optional_access() optional:54
std::optional::value() optional:404
openblack::DefaultWorldCameraModel::UpdateModeArcBall(vec<…>, vec<…>, float) DefaultWorldCameraModel.cpp:381
openblack::DefaultWorldCameraModel::UpdateMode(const openblack::Camera &, vec<…>, float, vec<…>, float) DefaultWorldCameraModel.cpp:316
openblack::DefaultWorldCameraModel::Update(duration<…>, const openblack::Camera &) DefaultWorldCameraModel.cpp:594
openblack::Camera::Update(duration<…>) Camera.cpp:211
openblack::Game::Update() Game.cpp:375
openblack::Game::Run() Game.cpp:802
main(int, char **) main.cpp:231
invoke_main() 0x00007ff6966f5c99
__scrt_common_main_seh() 0x00007ff6966f5b7e
__scrt_common_main() 0x00007ff6966f5a3e
mainCRTStartup(void *) 0x00007ff6966f5d2e
<unknown> 0x00007ffc06977344
<unknown> 0x00007ffc084c26b1
```

FalseIlyu:
I made some more checking, it isn't when you try to arcball while flying-by it is when you try to arcball with the mouse in the sky (_screenSpaceMouseRaycastHit becomes nullpot and so _screenSpaceMouseRaycastHitAtClick becomes nullopt).

test/camera/test_camera.cpp
```diff
@@ -0,0 +1,444 @@
...
+class TestDefaultCameraModel: public testing::TestWithParam<TestValues>
+{
+protected:
+	void SetUp() override
+	{
+		const auto testName = GetParam().name;
+		const auto testResultsPath = std::filesystem::path(k_ScenarioPath) / (testName + std::string(".json"));
+		if (!std::filesystem::exists(testResultsPath))
+		{
+			GTEST_SKIP() << "Unable to find the scenario test json, is it in the path?";
```
bwrsandman, line 58: Perhaps we should either have the json exported with the test or include them in the binary

bwrsandman:
@FalseIlyu I think I caught the bad optional accesses.

src/Camera/DefaultWorldCameraModel.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+			_targetOrigin =
+			    _targetFocus + glm::normalize(_targetOrigin - _targetFocus) * _originFocusDistanceAtInteractionStart;
+		}
+	}
+}
+
+void DefaultWorldCameraModel::UpdateModeFlying(glm::vec3 eulerAngles)
+{
+	if (!_handPosition.has_value() || !_screenSpaceCenterRaycastHit.has_value())
```
FalseIlyu, line 442: I tested it and didn't had any more problem with bad optional access.
```suggestion
	if (!_screenSpaceCenterRaycastHit.has_value() || !_screenSpaceMouseRaycastHitAtClick.has_value())
```
It isn't necessary to avoid bad access, but when double clicking on the sky now you fly toward some coordinates behind the camera so i think it would be better to ensure that the mouse hit the land before initiating a fly-by.

FalseIlyu (changes requested):

FalseIlyu:
I also tried to launch the camera unit-test again and they are running and passing so no more problem on my end with that.

src/Camera/DefaultWorldCameraModel.cpp
bwrsandman: oh yeah good catch

FalseIlyu (approved):

src/Camera/DefaultWorldCameraModel.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+			_targetOrigin =
+			    _targetFocus + glm::normalize(_targetOrigin - _targetFocus) * _originFocusDistanceAtInteractionStart;
+		}
+	}
+}
+
+void DefaultWorldCameraModel::UpdateModeFlying(glm::vec3 eulerAngles)
+{
+	if (!_handPosition.has_value() && !_screenSpaceCenterRaycastHit.has_value())
```
FalseIlyu, line 442:
```suggestion
void DefaultWorldCameraModel::UpdateModeFlying(glm::vec3 eulerAngles)
{
	if (!_screenSpaceMouseRaycastHitAtClick.has_value() || !_screenSpaceCenterRaycastHit.has_value())
```
I'm afraid you overwrote your previous correction for the Flying mode

FalseIlyu (changes requested):

src/Camera/DefaultWorldCameraModel.cpp
bwrsandman: I think you're mistaken, the other correction with `_screenSpaceMouseRaycastHitAtClick` is in `UpdateModeArcBall`

FalseIlyu: Sorry, i have mixed the bad access correction and my previous suggestion in this review.
```suggestion
void DefaultWorldCameraModel::UpdateModeFlying(glm::vec3 eulerAngles)
{
	if (!_handPosition.has_value() || !_screenSpaceCenterRaycastHit.has_value())
```
You had change the flyingMode to this in order to avoid bad access when double clicking on the sky (it is like this in my previous review).

The `_screenSpaceMouseRaycastHitAtClick` is my suggestion to avoid going at location behind the camera when double clicking on the sky, but i tested it more and it doesn't actually work so we may just drop that and keep in mind to change the raycast so it doesn't give position behind the camera.

bwrsandman: This suggestion makes line 446 dead code. Perhaps point should be replaced by _handPosition

FalseIlyu: It would probably best to apply the last suggestion and replace point by _handPosition as in the current state i do manage to get bad optional access. I'm trying to look into it further but you do access `_screenSpaceCenterRaycastHit` to compute `distanceFromHitPoint` and if `_handPosition` had a value it is possible for `_screenSpaceCenterRaycastHit` to not have one.

bwrsandman: The issue is that we don't have a way of setting `_handPosition` to `std::nullopt` when the mouse is on the sky.

bwrsandman: Something is very wrong with the `glm::intersectRayPlane` in `Game::Update` for the water plane.

bwrsandman: I'm thinking maybe we open a bug about this and have it be an open task.

FalseIlyu: Agreed

FalseIlyu (approved):

raffclar:
Can we increase the default movement speed for keyboard arrow input?

raffclar (approved):

bwrsandman:
@raffclar this could be done by modifying `k_InteractionSpeedMultiplier`, however, I think that will break the tests of the vanilla recordings.

What I would like is to have some or all of the constants be tunable in settings. #708

PR #729, opened 2024-05-06
Camera issues

FalseIlyu:
Solves a collection of issues found while testing the new camera that didn't warrant locking the PR or weren't found.

In order \:
- There where an invalid access to optional that was not caught during test (In my test i was on release so assert on the * operator were not checked).
- There where strange behavior when clicking on the sky.
- Arcballing ~and flying~ did not work on water.

So i \:
- Changed the check in `UpdateModeFlying` in `DefaultWorldCameraModel.cpp:442` so it check that `_screenSpaceCenterRaycastHit` and removed the check for `_handPosition` as it takes `_screenSpaceCenterRaycastHit` if empty.
- ~In `DeprojectScreenToWorld` at `camera.cpp:142` i used 1 instead of 0 as z value for `rayStartProjectionSpace`, as the depth buffer is reversed due to (#665).~
- I saved both the normal and Z reversed projection matrix in camera and use Z reverse for rendering and normal projection for raycasting.
- ~In `UpdateRaycastHitPoints` at `DefaultWorldCameraModel.cpp:527` i allowed the `RaycastMouseToLand` and `RaycastScreenCoordToLand` to take into account water using the parameter boolean.~
- In `UpdateFocusPointInteractionParameters` at `DefaultWorldCameraModel.cpp:248` i used `RaycastMouseToLand` including the water to compute `_screenSpaceMouseRaycastHitAtClick`

I know there where an issue how slow zooming was with the mouse wheel, but i don't want to have to decide what is a confortable zooming speed.

### DIFF
--- a/src/Camera/Camera.cpp
+++ b/src/Camera/Camera.cpp
@@ -12,6 +12,7 @@
 #include <glm/gtc/matrix_transform.hpp>
 #include <glm/gtx/euler_angles.hpp>
 #include <glm/gtx/intersect.hpp>
+#include <glm/gtx/transform.hpp>
 #include <glm/gtx/vec_swizzle.hpp>
 
 #include "3D/LandIsland.h"
@@ -98,11 +99,8 @@ Camera& Camera::SetProjectionMatrixPerspective(float xFov, float aspect, float n
 {
 	_xFov = glm::radians(xFov);
 	const float yFov = (glm::atan(glm::tan(_xFov / 2.0f)) / aspect) * 2.0f;
-	const float h = 1.0f / glm::tan(yFov * 0.5f);
-	const float w = h / aspect;
-	const float a = nearClip / (farClip - nearClip);
-	const float b = (nearClip * farClip) / (farClip - nearClip);
-	_projectionMatrix = glm::mat4x4(w, 0.f, 0.f, 0.f, 0.f, h, 0.f, 0.f, 0.f, 0.f, a, 1.f, 0.f, 0.f, b, 0.f);
+	// Inverse near and far for reverse z, we need to translate z by 1 to get back to the [0 1] range
+	_projectionMatrix = glm::translate(glm::vec3(0.f, 0.f, 1.f)) * glm::perspective(yFov, aspect, farClip, nearClip);
 
 	return *this;
 }
--- a/src/Camera/Camera.cpp
+++ b/src/Camera/Camera.cpp
@@ -149,7 +149,7 @@ void Camera::DeprojectScreenToWorld(glm::vec2 screenCoord, glm::vec3& outWorldOr
 	// projection space (z=1 is near, z=0 is far - this gives us better
 	// precision) To get the direction of the ray trace we need to use any z
 	// between the near and the far plane, so let's use (mousex, mousey, 0.5)
-	const glm::vec4 rayStartProjectionSpace = glm::vec4(screenSpaceX, screenSpaceY, 1.0f, 1.0f);
+	const glm::vec4 rayStartProjectionSpace = glm::vec4(screenSpaceX, screenSpaceY, 0.999f, 1.0f);
 	const glm::vec4 rayEndProjectionSpace = glm::vec4(screenSpaceX, screenSpaceY, 0.5f, 1.0f);
 
 	// Calculate our inverse view projection matrix
--- a/src/Camera/Camera.cpp
+++ b/src/Camera/Camera.cpp
@@ -12,7 +12,6 @@
 #include <glm/gtc/matrix_transform.hpp>
 #include <glm/gtx/euler_angles.hpp>
 #include <glm/gtx/intersect.hpp>
-#include <glm/gtx/transform.hpp>
 #include <glm/gtx/vec_swizzle.hpp>
 
 #include "3D/LandIsland.h"
@@ -28,7 +27,8 @@ using namespace openblack;
 namespace
 {
 constexpr auto k_DefaultCameraOriginOffset = glm::vec3(0.0f, 0.0f, 120.0f);
-}
+constexpr auto k_ReverseZMatrix = glm::mat4(1.f, 0.f, 0.f, 0.f, 0.f, 1.f, 0.f, 0.f, 0.f, 0.f, -1.f, 0.f, 0.f, 0.f, 1.f, 1.f);
+} // namespace
 
 Camera::Camera(glm::vec3 focus)
     // Maybe a struct is not the right thing... Maybe an optional?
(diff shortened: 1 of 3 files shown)


src/Camera/Camera.cpp
```diff
@@ -98,11 +98,8 @@ Camera& Camera::SetProjectionMatrixPerspective(float xFov, float aspect, float n
 {
 	_xFov = glm::radians(xFov);
 	const float yFov = (glm::atan(glm::tan(_xFov / 2.0f)) / aspect) * 2.0f;
-	const float h = 1.0f / glm::tan(yFov * 0.5f);
-	const float w = h / aspect;
-	const float a = nearClip / (farClip - nearClip);
-	const float b = (nearClip * farClip) / (farClip - nearClip);
-	_projectionMatrix = glm::mat4x4(w, 0.f, 0.f, 0.f, 0.f, h, 0.f, 0.f, 0.f, 0.f, a, 1.f, 0.f, 0.f, b, 0.f);
+	_projectionMatrix = glm::mat4x4(1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0, 1.0, 1.0);
+	_projectionMatrix *= glm::perspective(yFov, aspect, nearClip, farClip);
```
bwrsandman, line 102: Do these 2 lines match any of the perspective functions here?
https://github.com/g-truc/glm/blob/45008b225e28eb700fa0f7d3ff69b7c1db94fadf/glm/ext/matrix_clip_space.hpp#L215

I believe there's a define that lets you select the right kind of perspective

src/Camera/Camera.cpp
```diff
@@ -146,10 +143,10 @@ void Camera::DeprojectScreenToWorld(glm::vec2 screenCoord, glm::vec3& outWorldOr
 	const float screenSpaceY = ((1.0f - screenCoord.y) - 0.5f) * 2.0f;
 
 	// The start of the ray trace is defined to be at mousex,mousey,1 in
-	// projection space (z=0 is near, z=1 is far - this gives us better
+	// projection space (z=1 is near, z=0 is far - this gives us better
 	// precision) To get the direction of the ray trace we need to use any z
 	// between the near and the far plane, so let's use (mousex, mousey, 0.5)
-	const glm::vec4 rayStartProjectionSpace = glm::vec4(screenSpaceX, screenSpaceY, 0.0f, 1.0f);
+	const glm::vec4 rayStartProjectionSpace = glm::vec4(screenSpaceX, screenSpaceY, 0.9f, 1.0f);
 	const glm::vec4 rayEndProjectionSpace = glm::vec4(screenSpaceX, screenSpaceY, 0.5f, 1.0f);
```
bwrsandman, line 150: I have a hard time understanding this code and why 1.0f doesn't work.
Is there anything to do with the end value? Maybe you have to be in the negatives?

bwrsandman (changes requested):

src/Camera/Camera.cpp
FalseIlyu: Afaik you can only choose whether you use OpenGL or Direct3D depth definitions (via define or using the _ZO or _NO functions respectively) the rest is choice between Left handed or right handed coordinates and what parameter to build the perspective with (fov / viewport). I don't think there is any built-in way to make a perspective matrix with a reversed z in glm.

bwrsandman: I'm trying to better understand the matrix here. I have reduced it to this. Note the inversing of far and near...
I don't quite understand the need for translation though...
```suggestion
	_projectionMatrix = glm::translate(glm::vec3(0.0f, 0.0f, 1.0f)) * glm::perspective(yFov, aspect, farClip, nearClip);
```

FalseIlyu: Because if you just reverse the far and nearPlane you remap the depth between -1 and 0.

src/Camera/Camera.cpp
FalseIlyu: Test pass at a value of 0.999 and right now i had test pass at 0.9995 so it seems we can approach one but if we are too close it starts to fail, maybe there is something i miss that mean we can't really reach 1 or it is a precision issue.

FalseIlyu: I did some test selecting and teleporting them and it seems fine, i have some trouble selecting the villager but i had problem with that before, moving by fly-by seems also fine even with a start ray at z =1.f. In game having z=0.999f doesn't seems to change much but it passes the unit test and 1. don't.

bwrsandman: Could be that you're running into a Vanilla bug that we've enshrined in a unit test unfortunately

src/Camera/Camera.cpp
```diff
@@ -98,11 +98,9 @@ Camera& Camera::SetProjectionMatrixPerspective(float xFov, float aspect, float n
...
 	_xFov = glm::radians(xFov);
 	const float yFov = (glm::atan(glm::tan(_xFov / 2.0f)) / aspect) * 2.0f;
-	const float h = 1.0f / glm::tan(yFov * 0.5f);
-	const float w = h / aspect;
-	const float a = nearClip / (farClip - nearClip);
-	const float b = (nearClip * farClip) / (farClip - nearClip);
-	_projectionMatrix = glm::mat4x4(w, 0.f, 0.f, 0.f, 0.f, h, 0.f, 0.f, 0.f, 0.f, a, 1.f, 0.f, 0.f, b, 0.f);
+	// Inverse near and far for reverse z, we need to translate z by 1 to get back to the [0 1] range
+	_projectionMatrix =
+	    glm::translate(glm::mat4(1.f), glm::vec3(0.f, 0.f, 1.f)) * glm::perspective(yFov, aspect, farClip, nearClip);
```
bwrsandman, line 103: There's a gtx glm header for translation that doesn't require the matrix .

glm/gtx/transform.hpp

https://glm.g-truc.net/0.9.0/api/a00192.html#ac0a389d3a4a74010983db8bccab78d7f

bwrsandman: Or you can translate the projection

src/Camera/Camera.cpp
FalseIlyu: I tried to get value for `DeprojectScreenToWorld` on frame 1 (fail at the start of frame 2) of MoveBackwardForward scenario while not reversing the perspective. This is when updating `_screenSpaceMouseRaycastHit`

Z = 1.f (Fail)
[998.717773 2.04257107 1118.00208] Origin
[997.435729 1.08520007 1116.00415] End
[-0.500863254 -0.37402162 -0.780540705] Direction

Z = 0.999f (Pass)
[998.716491 2.04161239 1118.00012] Origin
[997.435729 1.08520007 1116.00415] End
[-0.500856638 -0.374016047 -0.780547678] Direction

Normal Z = 0.f (Pass)
[998.717956 2.04264212 1118.0022] Origin
[997.436096 1.08534205 1116.00439] End
[-0.500833511 -0.374025106 -0.780558168] Direction

FalseIlyu:
I prefer having it noted in text here, but i added to the camera two perspective matrix one with reverseZ and one without.
Raycast computation are done with the normal perspective, rendering are done with the reversed one. I added a enum to decide which camera mode we would in getPerspectiveMatrix and think this stuff could be used as an option at a later date (just would need to figure out the Z-Fighthing for normal perspective).

bwrsandman (approved):

PR #356, opened 2022-03-19
Camera bookmarks

tino113:
- [x] Depends on #354

Issue #348, opened 2022-02-03
Implement bookmarks and fly-to


bwrsandman:
Using CTRL+1 to CTRL+7, you can set bookmarks on the map where the hand is. It also takes into account he relative position of the camera.
https://user-images.githubusercontent.com/1013356/152371043-3de7946b-c261-4916-acaa-6d66047a743b.mp4

Pressing 1 to 7 will fly the camera towards that position.
https://user-images.githubusercontent.com/1013356/152371109-7b518b44-0b1a-470f-9262-09e8e2afd1fd.mp4

This requires adding rendering support for sprites to represent the bookmark number which are cut-through transparent billboards

https://user-images.githubusercontent.com/1013356/152371518-844dd5c9-6829-4a65-ad36-256fc7921750.mp4

The fly-to logic might share the same logic as #347

tino113:
fly-to looks to be the same, not sure how to render sprites will take some research, any help with this part would be appreciated.. otherwise happy for this and the other to be assigned to me

bwrsandman:
Bookmark sprites are in ./Data/Textures/misc0a.raw
![image](https://user-images.githubusercontent.com/1013356/155460351-27b6e5ad-22ce-4b1b-bc8a-5644f473df00.png)

bwrsandman:
Rendering and setting bookmarks implemented in #354 but flying to is not implemented yet

PR #737, opened 2024-05-11
Implement Fly to bookmarks

bwrsandman:
Exposed functionality to enqueue a fly-to with the camera model.
Added the camera origin to the saved data of a bookmark.

The way bookmarks work is that when you assign one, the hand position sets the position of the bookmark which will act as the camera's focus point. The camera's origin is also saved. Both these are used for fly-to when you restore the bookmark,
The original focus is lost.

Implements #348
Supersedes #356

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -34,6 +34,7 @@
 #include "Debug/Gui.h"
 #include "ECS/Archetypes/HandArchetype.h"
 #include "ECS/Archetypes/PlayerArchetype.h"
+#include "ECS/Components/CameraBookmark.h"
 #include "ECS/Components/Mobile.h"
 #include "ECS/Components/Transform.h"
 #include "ECS/Map.h"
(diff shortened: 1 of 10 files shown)


FalseIlyu (commented):
I can save the bookmarks but i can't restore the bookmarks.

src/Game.cpp
```diff
@@ -250,13 +251,20 @@ bool Game::ProcessEvents(const SDL_Event& event)
...
 			{
-				// TODO(#348): Fly-to the bookmark
+				const auto& entitiesRegistry = Locator::entitiesRegistry::value();
+				const size_t index = event.key.keysym.sym - SDLK_1;
+				const auto& bookmarkEntities = Locator::cameraBookmarkSystem::value().GetBookmarks();
+				[[assume(index < bookmarkEntities.size())]];
+				const auto& entity = bookmarkEntities.at(index);
+				const auto& [transform, bookmark] =
+				    entitiesRegistry.Get<ecs::components::Transform, ecs::components::CameraBookmark>(entity);
+				_camera->GetModel().SetFlight(bookmark.savedOrigin, transform.position);
```
FalseIlyu, line 267: Here there need to be something to make sure we don't make access on unset bookmarks (either index check use an optionnal on the saved Origin)

FalseIlyu (changes requested):
In game when processing the keys `event.key.keysym.sym` is needs to be `SDL_GetKeyFromScancode(event.key.keysym.scancode)` or there going to be problem on other layout (eg why it didn't work on my keyboard which is on azerty by default).

src/Game.cpp
```diff
@@ -250,13 +251,20 @@ bool Game::ProcessEvents(const SDL_Event& event)
 		case SDLK_8:
 			if ((event.key.keysym.mod & KMOD_CTRL) != 0)
 			{
-				auto handPosition = _handPose * glm::vec4(0.0f, 0.0f, 0.0f, 1.0f);
+				const auto handPosition = _handPose * glm::vec4(0.0f, 0.0f, 0.0f, 1.0f);
 				const auto index = static_cast<uint8_t>(event.key.keysym.sym - SDLK_1);
-				Locator::cameraBookmarkSystem::value().SetBookmark(index, handPosition);
+				Locator::cameraBookmarkSystem::value().SetBookmark(index, handPosition, _camera->GetOrigin());
 			}
 			else if (event.key.keysym.mod == 0)
```
FalseIlyu, line 258:
```suggestion
			else
```
This check should probably be removed as it blocks restoring bookmarks while num lock is active

FalseIlyu (changes requested):

FalseIlyu (approved):

PR #742, opened 2024-05-18
Update runners to Clang 18

FalseIlyu:
Update runners using Clang 17 to Clang 18 to support enumerate views.

- [ ] MacOS
- [ ] Ubuntu Cross Compile


bwrsandman:
You won't be able to get clang 18 on android before ndk 27 which is currently in beta but which will only be released around maybe September 2024

https://github.com/android/ndk/wiki#release-schedule

bwrsandman:
ndk 27 is out

bwrsandman (approved):

Issue #680, opened 2024-01-02
Remove use of exceptions


bwrsandman:
In general, game engines do not use exceptions due to runtime costs and the extra ops they add.

Our usage pattern is pretty ubiquitous so it will likely have to be done in steps.

Using [`std:: expected`](https://en.cppreference.com/w/cpp/utility/expected) is a good substitute for the kind of exception handing we have.

We should start adding `noexcept` to hotpath and root functions while making sure clang-tidy picks up errors.

Third party libraries and headers will have to be configured in their no-exeption modes.

Finally, adding the proper compiler options to turn off exceptions should be added in cmake.

A good place to start would be in the components and apps.

Requires #656

Issue #656, opened 2023-08-18
Update to C++23


bwrsandman:
Requires setting https://github.com/openblack/openblack/blob/f2447f6faf02aa278738854930c5c2637f8cba22/CMakeLists.txt#L38

Steps first:
- [x] Search all TODO's related to c++ version and implement with newer c++ features such as #522 
- [x] Revisit issues related to version compatibility
- [ ] Revisit complicated templates and simplify using `concepts` (such as script to functions)
- [x] Fix errors in build
- [x] Any github action compiler download or settings

FalseIlyu:
Current state of CI update

# Cross Compile
| Test | OS Version | Compiler Version | Pass |
|----|---------|-------------------|------|
| android-armeabi-v7a-vcpkg, ubuntu-latest | Ubuntu 22.04.3 LTS | C/C++ Clang 14.0.7 | :x: |
| android-arm64-v8a-vcpkg, ubuntu-latest | Ubuntu 22.04.3 LTS | C/C++ Clang 14.0.7 | :x: |
| android-x86-vcpkg, ubuntu-latest | Ubuntu 22.04.3 LTS | C/C++ Clang 14.0.7 | :x: |
| android-x86_64-vcpkg, ubuntu-latest | Ubuntu 22.04.3 LTS | C/C++ Clang 14.0.7 | :x: |
| x86-linux, ubuntu-latest, ninja-multi-vcpkg, sudo dpkg --add-architecture i386 | Ubuntu 22.04.3 LTS | GNU 11.4.0 | :x: |
| x86-windows, windows-latest, ninja-multi-vcpkg | Microsoft Windows Server 2022 10.0.20348 Datacenter | MSVC 19.38.33133.0 | ✔️ |
| x86-windows, windows-latest, ninja-multi-vcpkg | Microsoft Windows Server 2022 10.0.20348 Datacenter | MSVC 19.38.33133.0 | ✔️ |

bwrsandman:
I'm going to mark this as done and any other sub task can be made into a new task

PR #748, opened 2024-05-22
Started moving to C++23 #656 #680

Afr0:
Started removing exceptions and replacing them with std::expected.
openblacklib now has to be compiled with C++23


bwrsandman:
Are you doing anything differently from #741 ?

bwrsandman:
Please base your changes on master.
Also be aware that calling `value()` [can throw an exception](https://en.cppreference.com/w/cpp/utility/optional/value)

bwrsandman:
#741 was merged, do you mind rebasing your PR on his work now?
If possible do something to reduce the number of commits (over 100 now!) because it makes it very hard to review what are your changes.

Guide on squash rebase
https://www.youtube.com/watch?v=AWayLpQHJeE

PR #768, opened 2024-08-06
LHVM implementation and integration

Daniels118:
- Implemented the `ScriptLibrary` component (#7);
- implemented the `lhvmtool` app;
- the `LHVMFile` class can now load both `.chl` and `Script_*.sav` files;
- added `ScriptHeaders` folder with original C headers used by the CHL compiler and LHVM
- added the functions table to the `Game` class, with skeleton for all the 464 functions (POP of inputs and PUSH of outputs with the proper types)
- implemented some functions to allow the execution of basic CHL scripts
- improved the `LHVMViewer` class


src/ScriptHeaders/ScriptChallengeEnums.h
```diff
@@ -0,0 +1,151 @@
...
+ * Copyright (c) 2018-2024 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#ifndef __SCRIPTCHALLENGEENUMS_H__
+#define __SCRIPTCHALLENGEENUMS_H__
```
bwrsandman, line 11:
```suggestion
#pragma once
```

src/ScriptHeaders/ScriptChallengeEnums.h
```diff
@@ -0,0 +1,151 @@
...
+	CHALLENGE_GRADUATION,
+	CHALLENGE_BABY_SITTER,
+	CHALLENGE_OUTRO,
+	CHALLENGE_LAST,
+};
+
+//------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------
+
+#endif //__SCRIPTCHALLENGEENUMS_H__
```
bwrsandman, line 151:
```suggestion
```

src/ScriptHeaders/ScriptEnums.h
```diff
@@ -0,0 +1,260 @@
...
+ * Copyright (c) 2018-2024 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#ifndef __SCRIPTENUMS_H__
+#define __SCRIPTENUMS_H__
```
bwrsandman, line 11:
```suggestion
#pragma once
```

src/ScriptHeaders/ScriptEnums.h
```diff
@@ -0,0 +1,260 @@
...
+	SCRIPT_DEV_FUNCTION_CREATURE_HELP_ON,
+	SCRIPT_DEV_FUNCTION_CREATURE_HELP_OFF,
+	SCRIPT_DEV_FUNCTION_ENTER_SAVEGAMEROOM,
+
+	SCRIPT_DEV_FUNCTION_LAST,
+};
+
+//------------------------------------------------------------------------------------------------------------------------
+
+#endif //__SCRIPTENUMS_H__
```
bwrsandman, line 260:
```suggestion
```

bwrsandman (changes requested):
For enums, try to match the style in Enums.h
Mainly:
1. Use `#pragma once` guards
2. Use enum class for the namespacing
3. They should be in `namespace openblack`
4. A comment with a link to their provenance would be good
5. Use `CamelBack` case
6. Don't use c-style prefixes
7. Remove the `// ----` separators

If enums are used as strings, include an array of string to map them. See k_AbodeNumberStrs

src/ScriptHeaders/ScriptEnums.h
```diff
@@ -0,0 +1,247 @@
...
+	Flowers,
+	FemaleCreature,
+	Last,
+};
+
+enum class Bool
+{
+	True = 1,
+	False = 0,
+};
```
bwrsandman, line 175: Is this needed?

src/ScriptHeaders/ScriptChallengeEnums.h
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#pragma once
+
+namespace openblack::script
+{
+
+enum class ChallengeEnums
```
bwrsandman, line 15: This enum kind of looks like it comes from game data files. Is this possible?

components/ScriptLibrary/include/LHVM.h
```diff
@@ -0,0 +1,178 @@
...
+#include <array>
+#include <filesystem>
+#include <functional>
+#include <map>
+#include <string>
+#include <vector>
+
+#include "LHVMFile.h"
+
+namespace openblack::LHVM
```
bwrsandman, line 23:
```suggestion
namespace openblack::lhvm
```

components/ScriptLibrary/include/LHVMFile.h
```diff
@@ -0,0 +1,140 @@
...
+#include <cstdint>
+
+#include <array>
+#include <filesystem>
+#include <string>
+#include <vector>
+
+#include "LHVMTypes.h"
+
+namespace openblack::LHVM
```
bwrsandman, line 21:
```suggestion
namespace openblack::lhvm
```

components/ScriptLibrary/include/LHVMFile.h
```diff
@@ -0,0 +1,140 @@
...
+	    , _data(data)
+	    , _hasStatus(true)
+	    , _stack(stack)
+	    , _variableValues(variableValues)
+	    , _tasks(tasks)
+	    , _ticks(ticks)
+	    , _currentLineNumber(currentLineNumber)
+	    , _highestTaskId(highestTaskId)
+	    , _highestScriptId(highestScriptId)
+	    , _executedInstructions(executedInstructions) {};
```
bwrsandman, line 111: Implementation should go into the `.cpp` file

components/ScriptLibrary/include/LHVMFile.h
```diff
@@ -0,0 +1,140 @@
...
+	bool LoadStatus(std::istream& stream);
+	bool LoadStack(std::istream& stream, VMStack& stack);
+	std::vector<VMVar> LoadVariableValues(std::istream& stream);
+	void LoadTasks(std::istream& stream);
+	VMTask LoadTask(std::istream& stream);
+	void LoadRuntimeInfo(std::istream& stream);
+
+public:
+	LHVMFile()
+	    : _header {.magic = {' ', ' ', ' ', ' '}, .version = LHVMVersion::BlackAndWhite} {};
```
bwrsandman, line 77: What does it mean to have `    ` (four space chars) as magic? You could also just inline specify them on line 26.
All that being said. I don't really see the usefulness to store this after reading it.

components/ScriptLibrary/include/LHVMFile.h
```diff
@@ -0,0 +1,140 @@
...
+	std::vector<VMVar> _variableValues;
+	std::vector<VMTask> _tasks;
+	uint32_t _ticks {0};
+	uint32_t _currentLineNumber {0};
+	uint32_t _highestTaskId {0};
+	uint32_t _highestScriptId {0};
+	uint32_t _executedInstructions {0};
+
+	/// Error handling
+	void Fail(const std::string& msg);
```
bwrsandman, line 57: Might be asking a bit much because you were shown other components which use `Fail` but since we're moving away from exceptions, would it be possible to design the API without exceptions.

Do discus with us before spending time on this ask.

components/ScriptLibrary/include/LHVMFile.h
```diff
@@ -0,0 +1,140 @@
...
+	    , _stack(stack)
+	    , _variableValues(variableValues)
+	    , _tasks(tasks)
+	    , _ticks(ticks)
+	    , _currentLineNumber(currentLineNumber)
+	    , _highestTaskId(highestTaskId)
+	    , _highestScriptId(highestScriptId)
+	    , _executedInstructions(executedInstructions) {};
+
+	~LHVMFile() = default;
```
bwrsandman, line 113: `= default` is also an implementation which would go into the `.cpp` file as
```c++
# In LVMFile.h
~LHVMFile();
# In LVMFile.cpp
LHVMFile::~LHVMFile() = default;
```

components/ScriptLibrary/include/LHVMTypes.h
```diff
@@ -0,0 +1,371 @@
...
+enum class ScriptType : uint32_t
+{
+	SCRIPT = 1,
+	HELP = 2,
+	CHALLENGE_HELP = 4,
+	TEMPLE_HELP = 8,
+	TEMPLE_SPECIAL = 16,
+	MULTIPLAYER_HELP = 64,
+
+	ALL = 0xFFFFFFFF
```
bwrsandman, line 132: Many of your enums are in C style ALLCAPS, is this a requirement for the lhvm language or can it be changed to CamelCase, like we have for all our other enums (Barring some exceptions)?

components/ScriptLibrary/src/LHVM.cpp
```diff
@@ -106,274 +59,1424 @@ struct imemstream: virtual membuf, std::istream
...
+	_opcodesImpl[28] = &LHVM::Opcode28BrkExcept;
+	_opcodesImpl[29] = &LHVM::Opcode29Swap;
+	_opcodesImpl[30] = &LHVM::Opcode30Line;
+}
 
 /// Error handling
 void LHVM::Fail(const std::string& msg)
 {
-	throw std::runtime_error("LLVM Error: " + msg + "\nFilename: " + _filename.string());
+	throw std::runtime_error("LHVM Error: " + msg);
```
bwrsandman, line 104: Removing the filename from the error make me think that we can do without throwing exceptions. Instead return codes with string details like prototyped in #774

components/ScriptLibrary/src/LHVM.cpp
```diff
@@ -106,274 +59,1424 @@ struct imemstream: virtual membuf, std::istream
...
+	{
+		return _currentTask->exceptionHandlerIps.at(_currentTask->exceptionHandlerIps.size() - index - 1);
+	}
+	return 0;
+}
 
-	for (int32_t i = 0; i < count; i++)
+void LHVM::PrintInstruction(const VMTask& task, const VMInstruction& instruction)
+{
+	// TODO(Daniels118): improve this
```
bwrsandman, line 659: How, why? Maybe make an issue and put details there, then reword as  `// TODO(#ISSUE_NUMBER)`

contributors.md
```diff
@@ -7,3 +7,4 @@ Includes all git commit authors. Aliases are GitHub user names.
 * Alexandre Berdeaux (FalseIlyu) <ilyulemail@gmail.com>
 * Reinhold Gschweicher (NeroBurner) <pyro4hell@gmail.com>
 * Michał Janiszewski (janisozaur) <janisozaur@gmail.com>
+* Daniele Lombardi (Daniels118)
```
bwrsandman, line 10: As an author it is important to be able to contact you other than discord. Do you have maybe a privacy aware alias?
Needs might be:
* Permission to re-use work.
* Change of license.
* Credit.
* Questions.

src/Game.h
```diff
@@ -184,6 +191,498 @@ class Game
...
+	void VmGameClearComputerPlayerActions();              // 454 GAME_CLEAR_COMPUTER_PLAYER_ACTIONS
+	void VmGetFirstInContainer();                         // 455 GET_FIRST_IN_CONTAINER
+	void VmGetNextInContainer();                          // 456 GET_NEXT_IN_CONTAINER
+	void VmGetTempleEntrancePosition();                   // 457 GET_TEMPLE_ENTRANCE_POSITION
+	void VmSaySoundEffectPlaying();                       // 458 SAY_SOUND_EFFECT_PLAYING
+	void VmSetHandDemoKeys();                             // 459 SET_HAND_DEMO_KEYS
+	void VmCanSkipTutorial();                             // 460 CAN_SKIP_TUTORIAL
+	void VmCanSkipCreatureTraining();                     // 461 CAN_SKIP_CREATURE_TRAINING
+	void VmIsKeepingOldCreature();                        // 462 IS_KEEPING_OLD_CREATURE
+	void VmCurrentProfileHasCreature();                   // 463 CURRENT_PROFILE_HAS_CREATURE
```
bwrsandman, line 684: This addition is a no-go for us.
The game class is already too monolithic and we're trying to get rid of the game class entirely.
This addition add too much coupling and complexity to the game class.
You may want to look at how we handle Scripts.

A better solution would be to move this to a [locator service pattern](https://gameprogrammingpatterns.com/service-locator.html). See Locator.h.

bwrsandman (changes requested):
I'm really impressed so far with this PR. It's high quality.
There are a few things to change though. The main one is the integration in Game.h

If you feel like this is too big much of a task, feel free to make one PR for the component and app and another for the openblack integration.

components/ScriptLibrary/include/LHVMTypes.h
Daniels118: I can rename them without side effects.

components/ScriptLibrary/include/LHVMFile.h
Daniels118: There isn't a real need to store it, but the filed is part of the `LHVMHeader` struct. We could get rid of the `LHVMHeader _header` field in the `LHVMFile` class and replace it with `LHVMVersion version` (at the moment we don't even need to store the version, but it could be useful if we will ever support future VM versions).
As you said, the magic string can be read and checked without storing, and also hardcoded in the write method (when we'll implement it).

components/ScriptLibrary/src/LHVM.cpp
Daniels118: Many opcodes have different meanings depending on both `mode` and `datatype` fields. The current implementation of this print method doesn't take into account all these differences, so different instructions may be shown with identical keywords and we can't really understand what they do (this applies only to this print method, the actual implementation is fully functional). We don't need this right now because the game functions table isn't fully implemented yet, so it would be impossible to run complex CHL scripts, but when it will be ready, we may need correct display of instructions in order to debug the VM and/or the functions implementation.

PR #777, opened 2024-08-14
Add a field of view slider to the Debug menu

raffclar:


bwrsandman (approved):
Small improvement: Always get the `aspect` (or `1.0f`) and display it as read-only and increase the FOV limit to 180.

Also, please provide a screenshot of the menu :)

PR #788, opened 2024-08-22
camera: Fix math error in xFov to yFov conversion

bwrsandman:
The conversion function gets the ratio of depth to width of the right triangle that the view frustum traces (`tan(xFov / 2)`, applies the aspect ratio, to turn it to vertical (` / aspect`), turns that back to the angle for the right triangle (`atan`) and back to fov (`x 2`).

Our current implementation converts the angle to fov before applying `atan`.

This causes a bad angle to be fed to the perspective matrix function and it can be seen that something is wrong by resizing the game windows in the horizontal. At a small enough size, the image flips which indicates a negative FOV which shouldn not be possible while the width is positive.

With these changes, we have 1-to-1 correspondence with vanilla.

image
image


raffclar (approved):
Amazing!

Issue #754, opened 2024-05-24
Expand on Config


bwrsandman:
- [ ] Decouple Config from Game
- [ ] Store start-up params in a config file

PR #789, opened 2024-08-23
game: load startup arguments from config.json

R2RT:
Priority for arguments is:

1. Command line specified
2. Loaded from config.json
3. Compile-time fallback value

https://github.com/openblack/openblack/issues/754

Note: when pushing I've spotted that there are several copies of nlohmann::json library in repo, why are they not vckpg-provided? I've used jsoncpp at random, but switch to nlohmann is no problem.


src/main.cpp
```diff
@@ -40,24 +41,24 @@ bool parseOptions(int argc, char** argv, openblack::Arguments& args, int& return
 	{
 		loggingSubsystems += std::string(", ") + system.data();
 	}
+	const auto config = StartupConfig(std::filesystem::current_path() / "config.json");
```
bwrsandman, line 44: Current path is a good idea, I also think we should store something globally for the user such as in the user config directory

Function name | Value on Linux | Value on Windows | Value on macOS
-- | -- | -- | --
config_dir | `$XDG_CONFIG_HOME`       or `$HOME/.config` | `{FOLDERID_RoamingAppData}` or `{FOLDERID_LocalAppData}` | `$HOME/Library/Application Support`

There's also android to consider.

bwrsandman:
> Note: when pushing I've spotted that there are several copies of nlohmann::json library in repo, why are they not vckpg-provided? I've used jsoncpp at random, but switch to nlohmann is no problem.

They were provided as header only and for gltf, but I guess it makes sense to use it from pacman (on arch) and vcpkg at this point.

R2RT:
> > Note: when pushing I've spotted that there are several copies of nlohmann::json library in repo, why are they not vckpg-provided? I've used jsoncpp at random, but switch to nlohmann is no problem.
> 
> They were provided as header only and for gltf, but I guess it makes sense to use it from pacman (on arch) and vcpkg at this point.

Given this and recent discussion about keeping tools as kinda separate codebases, it seems that adding yet-another-copy should be preferred solution for now. I've added the very same one into `external` folder.

bwrsandman:
You libcap problems are likely solved with https://github.com/openblack/openblack/pull/792

PR #790, opened 2024-08-26
clang-tidy: fix execution for MSVC

R2RT:
~~In short, first 2 commits refactors boilerplate to functions, what enables 3rd commit: fix for running clang-tidy in msvc environment.~~

~~1st commit is just small fix for `clang-cl` warning I've encountered.~~

clang-tidy get reenabled for msvc for tools and components without any changes.
Only in case of `openblack_lib`, `/EHsc` has to be manually added to `clang-tidy` invocation, see commit description for details.

I've not looked if or how to enable `clang-tidy` for msvc in CI.

Now I can get back to https://github.com/openblack/openblack/pull/789 without having to run full CI for simple linting :)

### DIFF
--- a/apps/anmtool/CMakeLists.txt
+++ b/apps/anmtool/CMakeLists.txt
@@ -8,8 +8,7 @@ target_compile_definitions(anmtool PRIVATE CXXOPTS_NO_EXCEPTIONS)
 target_link_libraries(anmtool PRIVATE cxxopts::cxxopts anm)
 
 if (OPENBLACK_CLANG_TIDY_CHECKS)
-  # FIXME(bwrsandman) MSVC is throwing false errors about exceptions being disabled
-  if (NOT MSVC AND CLANG_TIDY)
+  if (CLANG_TIDY)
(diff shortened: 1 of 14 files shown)


bwrsandman:
I know it's tempting to want to factor common code in `components` but the design idea behind them is to make it as easy as possible to extract.
This is why there some duplicated code. I know it seems like developer sin to do but it has not been a maintenance burden.
What we gain from this is independence of these components which can be used outside of the project with minimal effort.
Examples of usages are editors, blender plugins etc.

R2RT:
Sure thing, I did these refactors rather to document/show explicit intention of the code, given how much clang-tidy related code already takes. These are not this big components/tools (so far) to centralize common code for sake of maintanance. I had no idea about other uses for the code.

Anyway, I'll inline the changes.

PS. I was also trying to fix `cl : Command line warning D9025 : overriding '/EHs' with '/EHs-'`, but it would require more obscure solution manipulating `CMAKE_CXX_FLAGS`. So common code/refactor was even more reasonable.

bwrsandman (approved):
Thanks for taking a look at the FIXMEs and including the link to the issue.
I had forgotten what the problem had been

Issue #7, opened 2019-03-03
LHVM: Lionhead Virtual Machine (Challenge Scripts)


handsomematt:
Documentation on the `chl` file-format can be found here: https://github.com/openblack/openblack/blob/master/docs/chl.md

namgo:
I'm interested in getting involved with this project, and while the LHVM isn't an ideal place to start, it's something I find really cool. Could you say more about the status and goals for this issue?

I found the Challenge File Assembler and Challenge File Lexer that was released by one of the developers. I also saw the wiki page ( https://github.com/openblack/openblack/wiki/Challenge-File-(.chl) ) which references this. Additionally https://github.com/openblack/openblack/blob/883fd4978a31ba44cc7daff4f688a0ebb7066011/components/ScriptLibrary/docs/opcodes.md expands on this.

I'm still not quite sure about the status of integrating the VM into the game.

I'm on this project's discord now and would like to discuss this further.

-- I'm still figuring out the codebase, so I'm not quite sure what all does what, sorry if this question is just a misreading.

PR #791, opened 2024-08-30
Lhvm integration

Daniels118:
- Added *ScriptHeaders* folder with enums used by CHL scripts;
- Added the `CHLApi` class, which provides a functions table used to initialize the VM. It comes with the skeleton for all the 464 functions (POP of inputs and PUSH of outputs with the proper types). The `CHLApi` can be accessed through the `Locator` service.
- implemented some functions to allow the execution of basic CHL scripts (running other scripts, creating objects and moving them around).

Address issue https://github.com/openblack/openblack/issues/7


bwrsandman:
You libcap problems are likely solved with https://github.com/openblack/openblack/pull/792

src/ScriptHeaders/ScriptEnumstwo.h
```diff
@@ -0,0 +1,80 @@
+/******************************************************************************
```
bwrsandman, line 1: Why this filename and not just add these to Enum.h?

src/CHLApi.h
```diff
@@ -0,0 +1,44 @@
...
+
+class CHLApi
+{
+public:
+	CHLApi();
+
+	[[nodiscard]] const std::vector<lhvm::NativeFunction>& GetFunctionsTable() { return _functionsTable; };
+
+private:
+#define CREATE_FUNCTION_BINDING(NAME, STACKIN, STACKOUT, FUNCTION)       \
```
bwrsandman, line 30: You should move this to the .cpp

bwrsandman (changes requested):
Small changes but I like what I see. 

One thing though, I'd add logging for the to-do

src/ScriptHeaders/ScriptEnumstwo.h
Daniels118: There is no technical reason to have 2 distinct files, but the original scripting tools from LH have these 2, so keeping the same filenames make it easier to remember from where they came from. If you wish to maintain less files, I could merge them.

bwrsandman (approved):

PR #810, opened 2024-09-17
Add TextRenderer API which loads and renders Black & White fonts

handsomematt:
Adds a TextRenderer class

I will write more


bwrsandman:
Will have to add `utf8cpp-git` to the aur packages pre-installed for the AUR packaging.
Fix a memory leak
```
==3258== LEAK SUMMARY:
==3258==    definitely lost: 0 bytes in 0 blocks
==3258==    indirectly lost: 0 bytes in 0 blocks
==3258==      possibly lost: 335,544,320 bytes in 5 blocks
==3258==    still reachable: 0 bytes in 0 blocks
==3258==         suppressed: 0 bytes in 0 blocks
==3258== 
```
Then satisfy clang-tidy.

assets/shaders/fs_text.sc
```diff
@@ -0,0 +1,13 @@
...
+
+#include <bgfx_shader.sh>
+
+SAMPLER2D(s_tex, 0);
+
+void main()
+{
+	vec4 texel = texture2D(s_tex, v_texcoord0.xy);
+
+	// gl_FragColor = vec4(1, 0.7, 0.7, texel.a);
```
bwrsandman, line 11:
```suggestion
```

src/Debug/Gui.cpp
```diff
@@ -116,6 +117,7 @@ std::unique_ptr<DebugGuiInterface> DebugGuiInterface::Create(graphics::RenderPas
 	debugWindows.emplace_back(new PathFinding);
 	debugWindows.emplace_back(new Audio);
 	debugWindows.emplace_back(new TempleInterior);
+	debugWindows.emplace_back(new FontViewer);
```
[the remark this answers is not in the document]
bwrsandman: This is a new warning... Maybe disable it or use `new (std::nothrow) FontViewer`

src/Game.cpp
```diff
@@ -888,6 +889,10 @@ bool Game::Run() noexcept
 			}
 		}
 
+		{
+			Locator::textRenderer::value().Draw();
```
bwrsandman, line 893: Missing profiler scope

src/Graphics/RenderPass.h
```diff
@@ -24,6 +24,7 @@ enum class RenderPass : uint8_t
 	Main,
 	ImGui,
 	MeshViewer,
+	Gui,
```
bwrsandman, line 27: I would put Gui before `ImGui`

src/Locator.cpp
```diff
@@ -117,6 +118,7 @@ bool openblack::InitializeEngine(uint8_t rendererType, bool vsync) noexcept
 
 bool openblack::InitializeGame() noexcept
 {
+	Locator::textRenderer::emplace<TextRenderer>();
```
bwrsandman, line 121: This needs to be reset at `~Game`. I know it's horrible. But that's how it is right now.

bwrsandman (changes requested):

src/Graphics/TextRenderer.cpp
```diff
@@ -0,0 +1,395 @@
+/******************************************************************************
+ * Copyright (c) 2018-2024 openblack developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/openblack/openblack
+ *
+ * openblack is licensed under the GNU General Public License version 3.
+ *******************************************************************************/
+
+#define LOCATOR_IMPLEMENTATIONS
```
bwrsandman, line 10: You're not checking for this because you're not using an abstract base class. In this case it's not needed.

src/Common/Color.h
```diff
@@ -0,0 +1,67 @@
+/******************************************************************************
```
bwrsandman, line 1: My opinion is that having a class like this is OOP overkill. I would just have some constants header with 
```
using Color = uint32_t;
namespace colors
{
constexpr Color k_White = 0xFFFFFFFF;
constexpr Color k_Black = 0x000000FF;
constexpr Color k_Good = 0xEBEBB7FF
constexpr Color k_Evil = 0xFFB4B4FF;
}; // namespace colors
```
Use comments if the element ordering needs to be clarified.

bwrsandman: Are we going to be doing a 1 to 1 reimplantation of LH3DColor?

PR #812, opened 2024-09-19
Revert commits converting `LOCATOR_IMPLEMENTATIONS` warnings to messages

bwrsandman:
The `LOCATOR_IMPLEMENTATIONS` warnings were converted from warnings to `#pragma message`.
This is a potential risk to future development because the CI treats warnings as errors to prevent the accumulation of valid warnings without preventing local development from being able to see warnings and continue working.
In the case of `LOCATOR_IMPLEMENTATIONS`, it is imperative that the warning is not raised in the upstream branches because it would indicate illegal use of the services implementations instead of the abstract base classses.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -60,4 +60,5 @@ CheckOptions:
     value: _
   - key: readability-identifier-naming.ProtectedMemberPrefix
     value: _
-# WarningsAsErrors: '*'
+ExtraArgsBefore:
+  - -DLOCATOR_IMPLEMENTATIONS
--- a/.clangd
+++ b/.clangd
@@ -1,2 +1,2 @@
 CompileFlags:
-  Add: [-std:c++20]
+  Add: [-std:c++20, -DLOCATOR_IMPLEMENTATIONS ]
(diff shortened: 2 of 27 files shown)


handsomematt:
Can we make this a single define that checks for MSVC? Since the message is repeated too.

bwrsandman:
What's the issue with MSVC?

handsomematt:
It does not know what #warning is

https://stackoverflow.com/questions/171435/portability-of-warning-preprocessor-directive

bwrsandman:
It should. Warning has been standardized in C23
https://en.cppreference.com/w/c/preprocessor/error

handsomematt:
Could we have something like this:

```
#define CHECK_LOCATOR_IMPLEMENTATIONS() \
    #if !defined(LOCATOR_IMPLEMENTATIONS) \
        #ifndef _MSC_VER \
            #warning "Locator interface implementations should only be included in Locator.cpp, use interface instead." \
        #else \
            #pragma message("Locator interface implementations should only be included in Locator.cpp, use interface instead.") \
        #endif \
    #endif
```

handsomematt:
Shall we ignore clang-diagnostic-error on clang-tidy? Makes sense to me

bwrsandman:
You can't. It's a parsing error but we can add `LOCATOR_IMPLEMENTATIONS` as a define

handsomematt (approved):

PR #817, opened 2024-10-21
Add MinGW support. Fix some openblack_lib include/link issues

twidds:
Adding configure and build presets for MinGW with Ninja/Makefiles build systems.  It would be nice to just use the existing ninja-multi-vcpkg preset, but it seems to require knowing MinGW is being used before vcpkg toolchain initializes so that the vcpkg triplet can be set properly, and I didn't find an easy way to do that without just splitting out a new preset.

Removed unnecessary ${BUILD_ROOT_DIR} from the include and link directories for the openblack_lib target. ${BULLET_INCLUDE_DIRS} and ${BULLET_LIBRARY_DIRS} are absolute paths, not relative. GCC and MSVC don't seem to care that the paths are invalid, but MinGW won't generate the RC objects with them in there.  Since the project previously compiled with the wrong paths I wonder if these lines could just be removed entirely?

SDL2main linkage in openblack_lib causes linker errors in MinGW and removing it didn't seem to impact GCC/MSVC linking, so I don't think it's necessary.  The project doesn't seem to use SDL2main at all anyway.


bwrsandman:
Looks nice but you'll have to fix the formatting errors that have been commented.

PR #825, opened 2025-01-26
ci: Use new linux arm runner

bwrsandman:


bwrsandman:
There is a fail on Packaging Arch Linux due to vcpkg having an older version of bgfx. This will be addressed later.
The failures on the pull_request targets are just due to the fact that they use old test configuration. The push equivalents are all passing except for arch.

PR #827, opened 2025-01-27
Fix vcpkg for bgfx and sdl2

bwrsandman:


hartmark:
For knowledge, commits here fixes build on AUR-package on Arch Linux.

https://aur.archlinux.org/packages/openblack-git

bwrsandman:
It's crashing because of the custom fmt formater....

PR #854, opened 2026-01-23
Last test

lucasbazan:


bwrsandman:
Could you keep your work in a single PR and keep it draft?

Also if you are just exploring, might I suggest you do experiments on your own repos until you have something to show? 

Not to be rude but I'd rather not have 5 different probing PRs in two days if they're just experiments.

The code is open source and you have the right to fork.

You might have more luck with a scaled down version of our CI that doesn't spend 20 minutes compiling bgfx

lucasbazan:
Sorry about that @bwrsandman this is my last test. I'd like to run it on my own profile, but I wonder that I don't have more minutes to use in my GitHub actions, but this is  my last test, sorry again for disturbing you

PR #859, opened 2026-05-02
feat(creature): load vanilla desire weights into state

JackDanHollister:
## Summary

Load Black & White's species-specific creature desire increase times from `info.dat` into per-entity ECS state, without inventing cognition behaviour.

`CreatureMind` becomes data plumbing sourced from `InfoConstants::creatureDesireForType`, and `CreatureArchetype::Create` bakes a spawned creature's values into a `CreatureState` component. Nothing consumes the values yet: per-turn desire accumulation is deliberately left for a follow-up grounded on the decomp.

## Changes

- Replace `CreatureMind::dummy` with the 40 desire increase times for a creature species, read from `InfoConstants`.
- Size `CreatureDesireForType` by the vanilla table width (`k_CreatureDesireForTypeColumns = 17`, `float ByCreatureType[0x11]` in bw1-decomp) and map openblack's `CreatureType` to its column through one `constexpr VanillaCreatureTypeIndex()`.
- Name the vanilla `CREATURE_TYPE` field of each `GCreatureInfo` record (`vanillaCreatureType`, record offset `0x1e4`). Layout and record size are unchanged.
- Qualify cached mind resources by creature type so creatures that share a mind filename do not share the wrong species values.
- Assign `CreatureState` in `CreatureArchetype::Create` only when the mind is loaded, so the component's presence is the initialised state.
- Tests: a hand-written column table with per-column sentinel probes, invalid-load recovery, species-qualified caching, and archetype-level population and absence.

## Vanilla data provenance

`InfoFile` reads the `Info` block of `info.dat` straight into `InfoConstants`. `creatureDesireForType` is 40 records of 17 floats. bw1-decomp (`CreatureMentalDesire.h/.cpp`, openblack/bw1-decomp#277) declares them as `float ByCreatureType[0x11]` and indexes them with `info->CreatureType` in `CreatureDesires::Initialise`, where vanilla `CREATURE_TYPE` is Ape = 0, Cow = 1 … Gorilla = 16. openblack's `CreatureType` has `Unknown = 0` and vanilla's Ape as `GiantApe = 17`, hence the mapping helper. In a retail `info.dat` the 17 `GCreatureInfo` records carry `vanillaCreatureType` 0 … 16 in order, Ape … Gorilla.

No speculative constants or fallback values remain.

## Review feedback addressed

- Values come from `info.dat`; the hard-coded weights and cognition constants are gone.
- Locator access with its own profiler scope was implemented in July and is now moot: the per-tick system was initialisation in disguise and has been replaced by archetype-time population.
- `CreatureState` carries no initialisation flag; presence is the state.
- The name→type map was deleted; `CreateCreatureFromFile` already receives `CreatureType`.
- The column-order question is fixed and documented at the declaration, with the mapping checked against retail data.

## Validation

- `clang-format-19` and `git diff --check`: clean on all changed files.
- Debug build with Clang 19: passed.
- CTest: 8/8; `test_creature_mind`: 7 cases.
- Real Black & White data: `Land1.txt` for 60 frames and `Playgrounds/FourGods.txt` for 120 frames (three `CREATE_CREATURE_FROM_FILE` calls sharing one mind file across species), both exit 0.

## Collaboration disclosure

This implementation was developed collaboratively by Jack Hollister, Hermes Agent by Nous Research, OpenAI GPT-5.6-sol and OpenAI Codex (GPT-5.6) for the earlier revisions, and Claude Opus 5 (Claude Code) for the September 2026 revision. Jack directed the design, provenance and publication decisions.


src/Creature/CreatureMind.cpp
```diff
@@ -0,0 +1,56 @@
...
+	set(CreatureDesires::ToLookAround, 0.30f);
+
+	// Threat responses
+	set(CreatureDesires::Fear, 0.65f);
+	set(CreatureDesires::Anger, 0.40f);
+
+	// Miscellaneous
+	set(CreatureDesires::Sadness, 0.20f);
+	set(CreatureDesires::ToRest, 0.35f);
+	set(CreatureDesires::ToStayNearHome, 0.30f);
```
bwrsandman: These are magic values. How did you derive them? Usually these kind of values come from info.dat

src/Creature/CreatureMindSystem.h
```diff
@@ -0,0 +1,46 @@
...
+	[[nodiscard]] CreatureDesires Plan(const ecs::components::CreatureState& state) const;
+	void Act(const CreatureDesires desire, ecs::components::CreatureState& state,
+	         ecs::components::LivingAction& action) const;
+	void ApplyReinforcement(ecs::components::CreatureState& state) const;
+
+	static constexpr float k_DetectionRadius = 50.0f;
+	static constexpr float k_EnergyDrainPerTick = 0.0002f;
+	static constexpr float k_MaxRLDelta = 2.0f;
+	static constexpr float k_RLDecayPerTick = 0.0001f;
+	static constexpr float k_ReinforcementStrength = 0.05f;
```
bwrsandman: Can we get some information about where this is from?

src/Resources/Loaders.cpp
```diff
@@ -173,9 +176,52 @@ LevelLoader::result_type LevelLoader::operator()(FromDiskTag, const std::filesys
...
 }
 
-CreatureMindLoader::result_type CreatureMindLoader::operator()(FromDiskTag, const std::filesystem::path& /*unused*/) const
+CreatureMindLoader::result_type CreatureMindLoader::operator()(FromDiskTag, const std::filesystem::path& creatureMindPath) const
 {
-	return std::make_shared<creature::CreatureMind>();
+	auto mind = std::make_shared<creature::CreatureMind>();
+
+	// Map filename to CreatureType so we can pick the right column in InfoConstants.
+	static const std::unordered_map<std::string, CreatureType> k_NameToType {
```
bwrsandman: This should be generalized to a reusable place

src/Game.cpp
```diff
@@ -265,6 +265,7 @@ bool Game::GameLogicLoop() noexcept
 	{
 		auto actions = profiler.BeginScoped(Profiler::Stage::LivingActionUpdate);
 		Locator::livingActionSystem::value().Update();
+		_creatureMindSystem.Update();
```
bwrsandman: This is apart from the living actions I would put this in its own profiler scope.

Also. See how the systems are services in Locator. This is how access to the mind system should be designed.

src/Game.h
```diff
@@ -18,6 +18,7 @@
 #include <glm/mat4x4.hpp>
 #include <spdlog/common.h>
 
+#include "Creature/CreatureMindSystem.h"
```
bwrsandman: We want to phase out the game mega class and move the header includes and members to the locator

bwrsandman (changes requested):
I appreciate the work.

There are automation fails to fix. Perhaps the gradle error can be ignored for now.

Few things to fix to get in, there are open black design patterns that this PR will need to conform to.
Another this that's very important is to not hard code game values if they weren't hard-coded in vanilla. I see a lot of such values that I would expect to come from dat files.

Also if you've had AI assistance such as for analysing the runblack.exe, planning changes or writing code, please disclose this in your description and commit messages. Include the model version too.

src/InfoConstants.h
```diff
@@ -798,23 +798,9 @@ struct GTotemStatueInfo: GMultiMapFixedInfo
...
-	float field0x28;
-	float field0x2c;
-	float field0x30;
-	float field0x34;
-	float field0x38;
-	float field0x3c;
-	float field0x40;
+	// The parsed Black & White 1.20 structure stores 17 contiguous floats in CreatureType's on-disk order,
+	// from Cow through GiantApe.
+	std::array<float, static_cast<size_t>(CreatureType::_COUNT) - 1> byCreatureType;
```
bwrsandman: I don't understand why the desire for type would map to Cow through GiantApe.

src/ECS/Systems/Implementations/CreatureMindSystem.cpp
```diff
@@ -0,0 +1,49 @@
...
+			state.desireWeights.fill(0.0f);
+			state.initialized = false;
+			return;
+		}
+
+		const CreatureMind& mind = *mindManager.Handle(creature.mind);
+		state.desireWeights = mind.baseDesireWeights;
+		state.mind = creature.mind;
+		state.initialized = true;
+	});
```
bwrsandman: This is initialization code not update code. Seems like you put this in the wrong place.
I don't see any actual game turn update logic.

src/ECS/Components/CreatureState.h
```diff
@@ -0,0 +1,28 @@
...
+#include "Creature/CreatureMind.h"
+
+namespace openblack::ecs::components
+{
+
+struct CreatureState
+{
+	std::array<float, creature::k_DesireCount> desireWeights {};
+	entt::id_type mind {};
+	bool initialized = false;
```
bwrsandman: Components don't really need initialization states. Couldn't the existence of a component be enough to show their initialization state?

bwrsandman (changes requested):

bwrsandman:
I have a reference for you here for the vanilla info loading.
A lot of it is 100% byte matching.
Do you mind having a look and cross referencing with it?

bwrsandman:
My link for the reference wasn't posted. Have a look here: https://github.com/openblack/bw1-decomp/pull/277

JackDanHollister:
Sorry for the silence here — I did cross-reference against bw1-decomp #277 after your message but never said so. Answering the open threads now, and I've pushed three commits on top of your rebase:

- `3117eb9c` names the desire-table columns. `CreatureDesireForType` is now sized by `k_CreatureDesireForTypeColumns = 17` (the `float ByCreatureType[0x11]` width from `CreatureMentalDesire.h`), and the openblack→vanilla column mapping lives in one `constexpr VanillaCreatureTypeIndex()` next to the struct, pinned by `static_assert`s and a hand-written column table in the test rather than the mirrored expression I had before. I've also named the `uint32` at record offset `0x1e4` of `GCreatureInfo` as `vanillaCreatureType` — your `CreatureInfo::CreatureType`, `0x1f4` in memory behind the header that isn't on disk. Reading my retail `info.dat`, records 0–16 carry 0–16 in order, `Ape creature` through `Gorilla creature`, which is the value `CreatureDesires::Initialise` indexes with. So Ape is column 0 and openblack's `GiantApe` reads column 0. Record size is unchanged.
- `1cae6fe4` removes `CreatureMindSystem`, its Locator service and its profiler stage. You were right that it was initialisation inside an `Update()`; the retry branch was also unreachable, since `CreateCreatureFromFile` loads the mind before it creates the entity. `CreatureArchetype::Create` now bakes the increase times into `CreatureState` from the loaded resource and only assigns the component when the mind exists, so presence is the initialised state — no flag, no cached mind id. That unwinds the Locator wiring from May, but only because there's no longer a system to wire.
- `302076c7` fixes my email in `contributors.md`.

On "I don't see any actual game turn update logic": there isn't any yet, on purpose. This PR is the data plumbing — nothing consumes `CreatureState::desireIncreaseTimes` until desire accumulation lands, and I'd rather ground that on the decomp before writing it. Happy to fold it in here if you'd prefer it to land together, or keep this PR to the loading and follow up.

I've also rewritten the description, which still described the old system and the wrong "Cow through GiantApe" column order.

src/InfoConstants.h
JackDanHollister: You were right to query this — the comment was wrong. Vanilla indexes the 17 columns with `CREATURE_TYPE` (Ape = 0 … Gorilla = 16); openblack's enum has `Unknown = 0` and vanilla's Ape as `GiantApe = 17`. Fixed in `6c72a0f9`; `3117eb9c` sizes the array by the decomp width and moves the mapping into `VanillaCreatureTypeIndex` next to the struct. Checked against retail `info.dat` — see the main comment.

src/ECS/Systems/Implementations/CreatureMindSystem.cpp
JackDanHollister: Agreed — it was spawn-time initialisation. Removed in `1cae6fe4`; `CreatureArchetype::Create` populates `CreatureState` from the loaded mind. No per-turn logic in this PR by design (main comment).

src/ECS/Components/CreatureState.h
JackDanHollister: Done in `1cae6fe4` — `CreatureState` is just the array now and is only assigned when the mind was loaded, so its presence is the state.

src/Creature/CreatureMind.cpp
JackDanHollister: Gone — the values come from `InfoConstants::creatureDesireForType`, i.e. the `Info` block of `info.dat`, with no fallback.

src/Creature/CreatureMindSystem.h
JackDanHollister: Those constants were invented. The whole cognition layer was removed in the July revision.

src/Resources/Loaders.cpp
JackDanHollister: Deleted rather than generalised — `CreateCreatureFromFile` already receives `CreatureType`, so the loader takes the enum directly and the map was redundant.

src/Game.cpp
JackDanHollister: Done in July (own profiler scope, Locator service); now moot — the system was removed in `1cae6fe4`.

src/Game.h
JackDanHollister: `Game.h` is identical to master again.

src/InfoConstants.h
```diff
@@ -857,8 +854,13 @@ struct GSpeedThreshold
 
 struct GCreatureInfo: GLivingInfo
 {
-	std::array<uint8_t, 416> field0x1e4;
+	// Vanilla CREATURE_TYPE of this record (Ape = 0 ... Gorilla = 16), not openblack's CreatureType. bw1-decomp
+	// CreatureInfo::CreatureType, 0x1f4 in memory behind the 0x10-byte Base/GBaseInfo header that is not on disk.
+	uint32_t vanillaCreatureType;
```
bwrsandman, line 859:
```suggestion
	CreatureType creatureType;
```

Pointless comment. Writing the code in my suggested way is self documenting.
The use of vanilla shouldn't be in this file. Everything here is understood to be vanilla. We even have namespaces for specific vanilla versions.

src/InfoConstants.h
```diff
@@ -796,27 +796,24 @@ struct GTotemStatueInfo: GMultiMapFixedInfo
...
 };
 
+// Precondition: Cow <= type <= GiantApe; the caller rejects Unknown and _COUNT.
+[[nodiscard]] constexpr size_t VanillaCreatureTypeIndex(CreatureType type)
+{
+	return type == CreatureType::GiantApe ? 0 : static_cast<size_t>(type);
+}
+static_assert(VanillaCreatureTypeIndex(CreatureType::GiantApe) == 0);
+static_assert(VanillaCreatureTypeIndex(CreatureType::Gorilla) + 1 == k_CreatureDesireForTypeColumns);
+
```
bwrsandman, line 816:
```suggestion
```

If GiantApe isn't part of 1.2, then it should have its own namespace.
In any case, InfoConstants is an inappropriate place for a function.

src/InfoConstants.h
```diff
@@ -857,8 +854,13 @@ struct GSpeedThreshold
...
 struct GCreatureInfo: GLivingInfo
 {
-	std::array<uint8_t, 416> field0x1e4;
+	// Vanilla CREATURE_TYPE of this record (Ape = 0 ... Gorilla = 16), not openblack's CreatureType. bw1-decomp
+	// CreatureInfo::CreatureType, 0x1f4 in memory behind the 0x10-byte Base/GBaseInfo header that is not on disk.
+	uint32_t vanillaCreatureType;
+	std::array<uint8_t, 412> field0x1e8;
 };
+static_assert(sizeof(GLivingInfo) == 0x1e4);
+static_assert(sizeof(GCreatureInfo) == 0x384);
```
bwrsandman, line 863:
```suggestion
```

Redundant with `static_assert(sizeof(InfoConstants) == ...);

test/creature/test_creature_mind.cpp
```diff
@@ -0,0 +1,253 @@
+/******************************************************************************
```
bwrsandman, line 1: What are you testing here?
I don't think we want to support a test suite that tests pretty much nothing.

src/Creature/CreatureMind.h
```diff
@@ -9,14 +9,33 @@
 
 #pragma once
 
-#include <cstdint>
+#include <array>
```
bwrsandman, line 12: Would it make more sense for this to be a component?

bwrsandman (changes requested):

PR #861, opened 2026-07-05
Working on villagers pathing.

KunenDev:
Sofar Villagers will only run around inside their own towns.


KunenDev:
Accidental PR from my fork. was meant to sync, some of the changes is from playing around.

PR #866, opened 2026-08-03
feat(chl): implement SET_FOCUS object facing

JackDanHollister:
## Summary

Implement the missing CHL `SET_FOCUS` native by turning the target object toward a supplied horizontal world coordinate.

The original native metadata identifies function 34 as an object-and-coordinate operation, separate from the camera-focus natives. A Four Gods runtime audit encountered the missing function 46 times in 600 unpaused frames.

## Changes

- Pop the target coordinate and object from the VM stack.
- Follow `GScript::SetFocus` from bw1-decomp: warn on a null thing, error on a thing that cannot face a position, and leave script containers and creatures alone.
- Rotate other entities towards the target's horizontal position using openblack's pathfinding orientation convention, with no zero-distance guard, as in `Object::SetFocus`.
- Keep a wall-hugging entity's `WallHug::yAngle` in step with its transform.
- Coverage: native-table wiring, horizontal and diagonal facing (the latter pins the rotation sign), the unguarded zero-distance case, and creature, town and wall-hug entities.

## Scope

`Creature::SetFocus` (`BW1W120 0x004f6760`) is not decompiled yet and vanilla does not turn a creature on the spot, so creatures are logged and skipped rather than rotated. The Town/Flock/Dance member fan-out has no openblack equivalent yet. Camera behaviour is unchanged.

## Validation

- `clang-format-19` dry-run: passed.
- `git diff --check`: passed.
- Debug build: passed.
- CTest: 8/8 passed.
- Four Gods real-asset run: 600 unpaused frames, exit 0, with no remaining `SET_FOCUS` unimplemented errors.
- Interactive Land 1 real-asset run: exited 0 with no fatal or assertion errors.

## References

- [CHL native-function metadata](https://github.com/Daniels118/blackandwhite/blob/main/chlasm/src/it/ld/bw/chl/model/NativeFunction.java)
- [Object `SetFocus` declaration in bw1-decomp](https://github.com/openblack/bw1-decomp/blob/main/src/Black/Object.h)

## Collaboration disclosure

This change was developed collaboratively by Jack Hollister and OpenAI Codex (GPT-5.6) for the initial revision, and Claude Opus 5 (Claude Code) for the September 2026 revision.

### DIFF
--- a/src/CHLApi.cpp
+++ b/src/CHLApi.cpp
@@ -20,6 +20,10 @@
 #include <LHVMTypes.h>
 #include <entt/entity/entity.hpp>
 #include <entt/entity/fwd.hpp>
+#include <glm/geometric.hpp>
+#include <glm/gtc/constants.hpp>
+#include <glm/gtx/euler_angles.hpp>
+#include <glm/vec2.hpp>
 #include <glm/vec3.hpp>
 #include <spdlog/spdlog.h>
 
@@ -475,10 +479,29 @@ void MoveGameThing() // 033 MOVE_GAME_THING
 
 void SetFocus() // 034 SET_FOCUS
 {
-	// const auto position = PopVec();
-	// const auto object = Pop().uintVal;
-	// TODO(Daniels118): implement this
-	SPDLOG_LOGGER_ERROR(spdlog::get("scripting"), "CHLApi Function {}() not implemented.", __func__);
+	const auto position = PopVec();
+	const auto object = Pop().uintVal;
+
+	if (object == 0)
+	{
+		return;
+	}
+
+	auto& registry = Locator::entitiesRegistry::value();
+	auto* transform = registry.TryGet<Transform>(static_cast<entt::entity>(object));
+	if (transform == nullptr)
+	{
+		return;
+	}
+
+	const glm::vec2 direction(position.x - transform->position.x, position.z - transform->position.z);
+	if (glm::dot(direction, direction) == 0.0f)
+	{
+		return;
+	}
+
+	const auto angle = std::atan2(direction.y, direction.x);
+	transform->rotation = glm::eulerAngleY(-angle - glm::half_pi<float>());
 }
 
 void HasCameraArrived() // 035 HAS_CAMERA_ARRIVED
--- a/src/CHLApi.cpp
+++ b/src/CHLApi.cpp
@@ -31,7 +31,10 @@
 #include "3D/TempleInteriorInterface.h"
 #include "Camera/Camera.h"
 #include "ECS/Archetypes/MobileStaticArchetype.h"
+#include "ECS/Components/Creature.h"
+#include "ECS/Components/Town.h"
 #include "ECS/Components/Transform.h"
+#include "ECS/Components/WallHug.h"
 #include "ECS/Registry.h"
 #include "ECS/Systems/HandSystemInterface.h"
 #include "Enums.h"
@@ -45,7 +48,10 @@ using namespace openblack::ecs::archetypes;
 
 using openblack::Locator;
 using openblack::MobileStaticInfo;
+using openblack::ecs::components::Creature;
+using openblack::ecs::components::Town;
 using openblack::ecs::components::Transform;
+using openblack::ecs::components::WallHug;
 using openblack::ecs::systems::HandSystemInterface;
 using openblack::lhvm::DataType;
 using openblack::lhvm::VMValue;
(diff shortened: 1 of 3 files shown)


bwrsandman:
Tests are currently failing. Will be fixed with #868.

bwrsandman:
Have a look at this PR for byte exact decomp for the same functions you're re-implementing. https://github.com/openblack/bw1-decomp/pull/287

One thing you must be aware of is that not every Object sets focus the same way. For example the creature when set focus, it will not rotate directly.

JackDanHollister:
Thanks — I've read #287, and you were right that the branch was wrong for creatures rather than merely incomplete: it rotated them directly. Pushed `e1fbb34e` on top of your rebase:

- Mirrors `GScript::SetFocus`: warn on a null thing ("Thing no longer valid"), error on a thing that can't face a position, and no zero-distance guard — `LH3DMath::GetYAngle` has none, so `atan2(0, 0)` faces angle 0. The test that asserted the opposite is inverted.
- Creatures: no rotation is applied. Vanilla goes through the virtual `Creature::SetFocus` (`0x004f6760`), and that body isn't in the decomp yet (`Creature.cpp` is still just includes), so the branch logs and returns rather than guessing. From the declarations #287 added alongside it — `PrepareCreatureForScriptedAction`, `SubArgumentPoint`, `CREATURE_SUB_STATE_ACTIONS_7`, `LookAtPosition` — it looks like it queues a look-at sub-action rather than touching the angle. Is that the shape? If so, a creature action path needs to exist before this can be done properly.
- Script containers: a Town has a Transform here, so the old code would have rotated the town entity. It now logs and returns; the member fan-out isn't expressible in openblack yet.
- `WallHug::yAngle` is kept in step with the transform, standing in for `SetYAngle` → `MobileWallHug::SetYAngle` (also not decompiled, so only the angle is mirrored).
- A diagonal-target test now pins the sign of the rotation; the previous test used a direction where `atan2` was 0.

One thing I'd like your view on: vanilla does `SetYAngle(atan2(dz, dx))` with no offset, while the branch uses `eulerAngleY(-angle - π/2)` to match `PathfindingSystem::InitializeStep`, so a villager told to face a point faces it the same way it faces its walking direction. But the archetypes use `eulerAngleY(-angle)` without the π/2, so the tree isn't consistent with itself. If you know which one the meshes actually want, I'll follow it.

Also — this push has queued another set of workflow runs that need approving; though as far as I can tell the VCPKG runtime jobs can't fetch `BW_ASSET_TOKEN` from a fork PR since #868 regardless.

test/chlapi/test_set_focus.cpp
```diff
@@ -0,0 +1,195 @@
+/******************************************************************************
```
bwrsandman, line 1: I don't think this feature warrants its own test suite...

src/CHLApi.cpp
```diff
@@ -475,10 +485,53 @@ void MoveGameThing() // 033 MOVE_GAME_THING
...
+
+	const auto entity = static_cast<entt::entity>(object);
+	auto& registry = Locator::entitiesRegistry::value();
+	auto* transform = registry.TryGet<Transform>(entity);
+	// Vanilla rejects anything that is not an Object; without an Object class, a missing Transform is the nearest check
+	if (transform == nullptr)
+	{
+		SPDLOG_LOGGER_ERROR(spdlog::get("scripting"), "CHLApi Function {}(): Thing must be living to face position!", __func__);
+		return;
+	}
```
bwrsandman, line 505: We don't have a concept of Object yet. This is either a TODO + issue or to be skipped. A Transform is not a good substitute.

src/CHLApi.cpp
```diff
@@ -475,10 +485,53 @@ void MoveGameThing() // 033 MOVE_GAME_THING
...
+
+	// Vanilla applies SetFocus to every member of a script container (Town/Flock/Dance) instead of the container itself
+	if (registry.AllOf<Town>(entity))
```
bwrsandman, line 508: Should be a TODO for `Flock` and `Dance`.
Shouldn't it be `AnyOf` in anticipation of those components?

src/CHLApi.cpp
```diff
@@ -475,10 +485,53 @@ void MoveGameThing() // 033 MOVE_GAME_THING
...
+	{
+		SPDLOG_LOGGER_WARN(spdlog::get("scripting"), "CHLApi Function {}(): script container focus not implemented", __func__);
+		return;
+	}
+
+	// Vanilla dispatches creatures to Creature::SetFocus (BW1W120 0x004f6760), which does not turn them on the spot.
+	// That override is not decompiled yet, so apply nothing rather than the generic rotation
```
bwrsandman, line 515: Should be a TODO. A decomp offset and vanilla function name isn't appropriate here. This is a reimplementation not a 1 to 1 decomp.

src/CHLApi.cpp
```diff
@@ -475,10 +485,53 @@ void MoveGameThing() // 033 MOVE_GAME_THING
...
+	{
+		SPDLOG_LOGGER_WARN(spdlog::get("scripting"), "CHLApi Function {}(): creature focus not implemented", __func__);
+		return;
+	}
+
+	// Object::SetFocus: yaw from the horizontal offset with no zero-distance guard, atan2(0, 0) == 0
+	const glm::vec2 direction(position.x - transform->position.x, position.z - transform->position.z);
+	const auto angle = std::atan2(direction.y, direction.x);
+	// Same angle-to-rotation mapping as PathfindingSystem's InitializeStep, so facing a point matches walking towards it
+	transform->rotation = glm::eulerAngleY(-angle - glm::half_pi<float>());
```
bwrsandman, line 526: Utility function?

bwrsandman (changes requested):
We haven't made an Agents.md for this repo.

One this you need to know if you're doing agentic coding is that we want comments to be human made. LLM comments are often misleading, stale and often inappropriate.

In this PR there is far too much mention of vanilla, vanilla functions and vanilla PE offsets.

This is a re-implementation and not a decomp. While it can take inspiration, reference and duplicate, Code written here needs to make sense as a modern C++ implementation.

